PR #13, opened 2016-03-20
Correctly process returned message at send audio

Clevero:
I tried to get the returned message when I send an audio file with sendAudio. 

response after a successful transmission:

```
{
   "result":{
      "date":1458073077,
      "reply_to_message":{
         "date":1458073068,
         "chat":{
            "id":My_id,
            "type":"private",
            "first_name":"My_Name",
            "username":"My_Username"
         },
         "message_id":771,
         "from":{
            "id":My_Id,
            "first_name":"My_Name",
            "username":"My_Username"
         },
         "text":"https://www.youtube.com/watch?v=C0DPdy98e4c"
      },
      "chat":{
         "id":My_Id,
         "type":"private",
         "first_name":"My_Name",
         "username":"My_Username"
      },
      "message_id":772,
      "from":{
         "id":MyBot_Id,
         "first_name":"MyBot",
         "username":"MyBot"
      },
      "audio":{
         "duration":0,
         "file_id":"[returned_file_id]",
         "title":"TEST VIDEO",
         "file_size":91438
      }
   },
   "ok":true
}
```

Out of my research every successfull transmission has a "result" object.

Please correct me, if I understand anything not in a correct manner

### DIFF
--- a/src/main/java/org/telegram/telegrambots/api/methods/SendAudio.java
+++ b/src/main/java/org/telegram/telegrambots/api/methods/SendAudio.java
@@ -17,6 +17,9 @@ import org.telegram.telegrambots.api.objects.ReplyKeyboard;
 public class SendAudio {
     public static final String PATH = "sendaudio";
 
+	public static final String DURATION_FIELD = "duration";
+	private Integer duration; ///< Integer	Duration of the audio in seconds as defined by sender
+
     public static final String CHATID_FIELD = "chat_id";
     private String chatId; ///< Unique identifier for the chat to send the message to (or Username fro channels)
     public static final String AUDIO_FIELD = "audio";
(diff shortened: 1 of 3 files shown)


rubenlagus:
Hi @Clevero, 

couple of details before I merge:

1. Message object already has the code you added ([check](https://github.com/rubenlagus/TelegramBots/blob/dev/src/main/java/org/telegram/telegrambots/api/objects/Message.java#L122)), why did you add them again? Any reason or just a mistake?

2.  The answer when making any request to Telegram Server always has a "ok" field, so it is not necessary to check `jsonObject.has("ok")`. It can be removed.

Clevero:
Sorry for that :(

I have apparently overlooked this

Is the new code ok?

PR #15, opened 2016-03-22
Downloading files via GET request

swoopyy:
I don't know exactly why downloading files from the Telegram servers is not available in the current version, but this is quite useful for those who want to receive files from the users.


rubenlagus:
Will be added as soon as I merge with master

PR #21, opened 2016-04-02
Java 7 compatibility

aNNiMON:


rubenlagus:
When I add new code, I always use java8, so even if I merge this, I can't give any warranty that new code will be java7 compatible. Why do you still need java7 compatibility? maybe it may be worth to add it to a different branch instead of master?

aNNiMON:
My friend uses your library on server with Java 7. I agree with your point, but library doesn't use any Java 8 specific code (such as streams, `java.util.function` etc), there's only one lambda expression for now, so it would be great to support Java 7. Of course, if you plan to add something that cannot do in Java 7, it's ok to switch back to Java 8 later.

rubenlagus:
Ok, I'll make the code compatible for now with java 7, but I won't change pom itself. Maybe I can leave a branch with the java7 pom file if you need it ;)

rubenlagus:
Should be already merged in Merge #38

Issue #27, opened 2016-04-12
Enhancement: setters to builder style


Pixelase:
Hi!
Would it be appropriate to make setters in the builder style?
Like this: 
```java
    public SomeClass setField(String field) {
        this.field = field;
        return this;
    }
```
In my opinion, this will add a bit of usability.
```java
sendMessage(new SendMessage().setChatId(chatId).setText("Text"));
````
If you need it, i can handle it and send you pull request.
Regards.

rubenlagus:
Looks nice, not that I like to return always a pointer, but indeed it may be helpful in those cases. Feel free to go ahead with it ;)

rubenlagus:
But please, make the pull request agains dev branch ;)

Pixelase:
@rubenlagus ok ;)

PR #30, opened 2016-04-12
Setters and some methods to builder style

Pixelase:
Here is my pull request that i [promissed ](https://github.com/rubenlagus/TelegramBots/issues/27)to do :)
I divided changes into commits. So you can choose exactly what changes you want.

### DIFF
--- a/src/main/java/org/telegram/telegrambots/TelegramBotsApi.java
+++ b/src/main/java/org/telegram/telegrambots/TelegramBotsApi.java
@@ -125,30 +125,33 @@ public class TelegramBotsApi {
      *
      * @param bot
      */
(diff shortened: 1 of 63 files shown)


rubenlagus:
@Pixelase Great! I'll merge this asap. I'm merging everything but the TelegramBotsApi ones, I don't really like that in a "Manager" like that.

In addition, I'll be removing all unnecessary setters from object classes that are not supposed to be instantiated from "outside" (like Message, User, Chat, etc)

Pixelase:
@rubenlagus Good news! 
Yeah, you are right about setters in object classes that serve as models. Okay then.
Regards.

PR #39, opened 2016-04-19
fix_inline_response

tolwi:


pavel-kurnosov:
+1 to merge it and build new version - now inline query is not working at all.

PR #48, opened 2016-04-23
Add Contact getters/setters

antonu17:
I want to get contact's data when receiving contact


rubenlagus:
Hi @antonu17,

in that object, there is no need for setters as it is only "read-only" object that comes inside the Message obejct. Will merge with those fixes.

antonu17:
I can show you snippet:
I receiving contact (sent via request_contact button):

    public void processMessage(Message message) {

        Contact contact = message.getContact();

        if (contact != null) {
            if (!Objects.equals(message.getFrom().getId(), contact.getUserID())) {
                   // User sent wrong contact (not via request_contact button)
                   return;
            }

            String phoneNumber = contact.getPhoneNumber().replaceAll("[^0-9]", "");
            // work with phone number
            ...

antonu17:
Setters removed

antonu17:
What do you think about this
https://github.com/rubenlagus/TelegramBots/issues/56

antonu17:
There was exception on building response on sendDocument, sendPhoto, sendVideo, sendSticker and sendVoice:
```
org.json.JSONException: JSONObject["message_id"] not found.
        at org.json.JSONObject.get(JSONObject.java:473)
        at org.json.JSONObject.getInt(JSONObject.java:535)
        at org.telegram.telegrambots.api.objects.Message.<init>(Message.java:121)
        at org.telegram.telegrambots.bots.AbsSender.sendPhoto(AbsSender.java:497)
````

antonu17:
I think i did wrong, push everything in master and all commits are here. I have to split changes by several branch and PRs. If you are not agree with any of last commits - let me know, i'll remake it

antonu17:
Ok, i see remake is needed

PR #52, opened 2016-04-25
Changed the json cahed object class to Sticker

camilosampedro:
It was generating a `InlineQueryResultCachedGif` instead of `InlineQueryResultCachedSticker`, producing a wrong behavior.


rubenlagus:
Perfect, in any case, that part shouldn't be important (except in debugging). Merging, thanks

PR #65, opened 2016-05-05
Fix some methods such as sendPhoto

balkin:
Actually such methods as sendPhoto were broken (library should use the result object, not the wrapper object). I made sure sendPhoto works OK, but other methods such as sendVoice are untested


rubenlagus:
Duplicate #63, should be fixed when merging that pull request.

Issue #41, opened 2016-04-21
Command API


lfdajr:
Hello,

I'd like do suggest a command API to handle the commands issued in the platform in a decoupled way. It would work like this:

1. You register the commands your robot treat
2. For each command you create a class which handles that command accordingly
3. All the code will stay contained.

I have done this in my local copy, if you guys would like to see it, I could submit.

Regards.

rubenlagus:
@lfdajr It would be great to have a look to it. Do you have it in a git for or something?

Mit0x2:
See #76 where I suggested my version of the above mentioned 🙂

rubenlagus:
I guess this one can be closed too, thanks @Mit0x2

PR #76, opened 2016-05-20
Suggestion for Command mechanism

Mit0x2:
Hi there,

according to #41 I tried to find a way of implementing an comfortable way of using commands. I stumbled across the message properties, where telegram already mentions whether the message is a command or not. Further I added a command registry, which manages the commands of a bot. For further functionality a added a standard help command, which displays all registered commands to the requesting user.
Hope you can do something with this :) Would be great to see something of this manner in your java telegram bot API!

Best wishes
Mit0


rubenlagus:
Nice work, I'll check it. 

In any case, I have planned a new version for the end of this week or next week mostly, so I don't think I'll include those changes in that one. Better to leave it for later ;)

rubenlagus:
@Mit0x2 There is one important problem that I can't merge: the chatId can't be a long because for channels the chatId is the username -> string.

Regarding, the commands, I need to give a deeper look, but I find it a big overload to make them extend from AbsSender, maybe just adding it as a parameter of the constructor is more efficient?

Mit0x2:
Hi @rubenlagus thanks for having a look :)
I'l have a look into the chatId Stuff. I agree that having a Command implementing AbsSender is a bit of a overload. I quite like the idea of simply reaching it in via constructor or sth equivalent. Will change those things and report back!

antonu17:
We will have to rewrite all out bots, because onUpdateReceived() become final.

Mit0x2:
Hi,
as mentioned this is only a suggestion. Secondly this change would be rolled out in another version, so non of your bots would be unfunctional from one day to another.
On top you would have to change the library version in your pom, so you could as well simply change the method overwritten by your bot. You might then want to use the commands as well, so a simple method rename shouldn't hurt that much. I could have solced it differently, but i wanted the method names to make sense and since it is no full update any more, because commands are filtered, it would be malicious leaving all the methodnames like they were.

I agree, that the current change would break all current implementations of commands. Maybe I could move my Implementation to a new "CommandAwareTelegramLongPollingBot"?! But this name would be way to huge 😔

As always I'm open to all of your suggestions.

rubenlagus:
I agree that we can't force such a big change in the API like making "onUpdateReceived" final. I'm even building new bots that don't even have commands at all (but /start one). 

I really do like the idea of adding it as a different bot class. It could extend from "TelegramLongPollingBot", can't it? About the name, well it is something that you only declare once per bot, shouldn't be any issue to have something a bit longer... we can live with that I'd say: CommandAwareTelegramLongPollingBot or just CommandsTelegramLongPollingBot could do.

Mit0x2:
Done :) Guess it was a good idea, so people can choose whether they need commands or not!

Thanks for mentioning this @antonu17 ! And thanks again for having a look @rubenlagus !
Hope you guys like my fix for the issue

Kurpilyansky:
I think it is no need store AbsSender inside of BotCommand. It's better to pass AbsSender as a parameter of method execute.

Mit0x2:
Thought about the same thing! Now when i saw, that i'm not the only one thinking that way, i did it ;)
Quite like the side effect, that now you could also reach one instance of a command into multiple bots without sending messages over the wrong bot instance. Would probably have caused strange issues...

Another thing is, that now it's much cleaner, don't know, why i did it differently in the first place! Thanks for your opinion @Kurpilyansky !

Mit0x2:
@rubenlagus do you want me to close this pull request and reopen it against the dev branch?

Mit0x2:
Reopened against Dev in PR #83

PR #83, opened 2016-05-24
Implement command mechanism - Reopened against Dev

Mit0x2:
For previous discussion  see #76 . This Pull request suggests a possible implementation of a built in command handling - see #41


rubenlagus:
Now it looks nice from here. I'll merge it before next version and check a couple of things:

1. I guess the new *LongPollingCommandBot* can extends directly *LongPollingBot* and implement *ICommandsRegistry*. Also I may add a similar implementation for *WebhookBot*.
2. *isCommand* in *Message* object now check if there is, at least, one command in the message, but it may be more useful to check also if the command is at the beginning of the message (or even if the command is the full message). Will see

Thanks @Mit0x2

rubenlagus:
Hi @Mit0x2,

I got time to have a look to it. Finally I made a couple of changes to make it more general:

1. HelpCommand: I removed it as it may not be a default action in every bot (Better not to enforce any behaviour here). On the other hand, I liked the way of converting the commands to string in order to send them, so I added it as default `toString` behaviour of *BotCommand.java*. This way we allow to implement custom help command and also to override the toString if desired.
2. In Message.java, I changed the `isCommand` method so it enforces also to have the command at the beginning of the message (`offset == 0`). Can't see any case where a command wont be the first thing in a message, but if you do, we can move the `isCommand` method to the *BotCommand.java* and add a new parameter in the constructor like `boolean onlyFirstPosition` that will enable/disabled this behaviour for each individual command.
3. Instead of add a hardcoded `SendMessage` when a command is not found in the list of commands, I added a *defaultAction* in *CommandRegistry.java* that will allow to implement custom behaviour for such a case. If this `defaultConsumer` is not provided, nothing will be made with the message. In addition, when the Message doesn't match any command and there is no defaultAction, it is better that the library *understand* that the user want to handle the message in a custom way, so now it is also passed to `processNonCommandUpdate`.

I pushed those changes together with the merge to dev branch, can you have a look?

Thanks

PR #88, opened 2016-05-31
Command bot additions

Mit0x2:
Cleaned up some JavaDoc, which I missed on my CommandBot contribution...

Added example for new Command capabilities to readme.md

Cheers

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ In order to use Long Polling mode, just create your own bot extending `org.teleg
 If you like to use Webhook, extend `org.telegram.telegrambots.bots.TelegramWebhookBot`
 
 
-Once done, you just need to creat a `org.telegram.telegrambots.TelegramBotsApi`and register your bots:
+Once done, you just need to create a `org.telegram.telegrambots.TelegramBotsApi`and register your bots:
 
 ```java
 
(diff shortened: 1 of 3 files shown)


rubenlagus:
Maybe it makes sense to add a simple example in the samples project? Maybe a bot that just support:
- /start -> Welcome message
- /hello -> Send a "Hello world!" message
- Any other command -> Send a "Default command received" 
- Any other thing -> Send a "Not command found" 


What do you think? Won't that cover all cases easily?

Mit0x2:
Hi,
could do sth like that as well. I thought like this it would be all together. So people would't have to look into another project.
Just close this pull request if you like an example bot in your Example project better. I'll then reopen a pull request with the attached minor javadoc changes.

rubenlagus:
Yep, if you like the idea, we can close this one until we have the example one and then add a nice reference from https://github.com/rubenlagus/TelegramBots#example-bots toguether with the other fixes you added in the docs (something that indeed needs a lot of work.. I wish I had time to write documentation...)

Mit0x2:
> something that indeed needs a lot of work.. I wish I had time to write documentation...
Will help out there in the near future. Probably will hijack the telegram group for questions :P

Will make the example later and add a pull request like mentioned.

PR #93, opened 2016-06-03
Feature/improve send newfile

### LINKED ISSUE #92: Change SendPhoto (and same) method for sending new file
Its not intuitive for new users to understand what they need to put in `setNewPhoto(String photo, String photoName)`. What is the second argument `photoName`? [Telegram Bot API sendPhoto documentation](https://core.telegram.org/bots/api#sendphoto) tells us nothing about `photoName`. And same kind of questions for same methods: sendDocument, sendVideo, etc.

antonu17:
Resolves #92 
I understand that it will change current API (at least a signature of `SendPhoto.setNewPhoto()` method), but it is for good.


antonu17:
Old methods reverted back, deprecated.
Add new methods, using InputStream (thanks to [@flicus](https://telegram.me/flicus) for sugestion using streams)

PR #112, opened 2016-07-13
Message filter for TelegramLongPollingCommandBot

dartwata:
Message filter.
Override this function in your bot implementation to filter messages with commands. For example, if you want to prevent commands execution incoming from group chat...


rubenlagus:
Was made agains master, reverted in #113 and merged to dev in 759d829252a2027fbc0647139dbe40aea0c8f001

PR #116, opened 2016-07-21
In some specific cases you got a StringIndexOutOfBoundsException

spolti:
IMO the length should be the message's length.
SEVERE: BOTSESSION
java.lang.StringIndexOutOfBoundsException: String index out of range: 66
  at java.lang.String.substring(String.java:1963)
  at org.telegram.telegrambots.api.objects.MessageEntity.computeText(MessageEntity.java:96)
  at org.telegram.telegrambots.api.objects.Message.lambda$new$0(Message.java:264)
  at java.util.ArrayList.forEach(ArrayList.java:1249)
  at org.telegram.telegrambots.api.objects.Message.<init😠Message.java:264)
  at org.telegram.telegrambots.api.objects.Update.<init😠Update.java:50)
  at org.telegram.telegrambots.updatesreceivers.BotSession$ReaderThread.run(BotSession.java:124)

To reproduce just use this message:
`JOptionPane jop = null;
null.showMessageDialog("bla bla bla");`

### DIFF
--- a/src/main/java/org/telegram/telegrambots/api/objects/MessageEntity.java
+++ b/src/main/java/org/telegram/telegrambots/api/objects/MessageEntity.java
@@ -93,7 +93,7 @@ public class MessageEntity implements IBotApiObject {
     }
 
     protected void computeText(String message) {
-        text = message.substring(offset, offset + length);
+        text = message.substring(offset, message.length());
     }
 
     @Override


rubenlagus:
The code you propose to reproduce the issue is not part of this library (as long as know). Also, I don't think the length must be the message one.. why do you think so?

ataxexe:
I think you've been mislead by the `JOptionPane jop = null; null.showMessageDialog("bla bla bla");`.

It was actually the content of the message sent through Telegram.

ataxexe:
The raw message is:

> `JOptionPane jop = null;

> null.showMessageDialog("bla bla bla");`

spolti:
The " `JOptionPane jop = null; null.showMessageDialog("bla bla bla");`" was the message used to reproduce this issue (pasted in the telegram chat), not the change itself, (note you need to use it with graves).

For some reason your api (or the telegram api itself) is taking of some character, idk which one exactly.
Once it happens, the message get stucked in the api because that method is not able to process it and the bot stops to work because of this.

rubenlagus:
Sending:

> \`JOptionPane jop = null; null.showMessageDialog("bla bla bla");`

Works for me, it generates this update correctly: http://pasteboard.co/egjoee3jp.png

ataxexe:
It's weird. We can't simulate it too.

Maybe it's related to another thing. But even so, I'd suggest you check the string bounds to prevent more exceptions like that.

rubenlagus:
Making that check will means that we are hiding the real issue: 

> Why are the offset and length parameters incorrect?

The best way would be getting a way to reproduce it, check if it is related to this project or not and, if not, add make a request to Telegram support so they fix it.


In any case, the solution in this merge request doesn't work in case you have more than one entities, so I'm closing it (feel free to open an issue instead).

spolti:
Very fun, it happened twice with two different bots, now it just doesn't happen anymore, hehe.

rubenlagus:
Maybe there was a bug indeed in Telegram side and they fixed meanwhile?

ataxexe:
Maybe. But even if was a bug. I'd suggest you check that substring. It's not safe to trust in that parameters because it can broke the bot entirely with an StringIndexOutOfBoundsException in that line (just like in our case).

PR #124, opened 2016-07-24
Update HOWTO

Clevero:
I hope this a HOWTO you are looking for. 

Since my English is not the best, I would appreciate if someone could go through and report some big mistakes :)

Also if it's wished I can add a HOWTO in german.


Clevero:
Should the [eclipse_configuration.md](https://github.com/rubenlagus/TelegramBots/blob/master/eclipse%20configuration.md) be mentioned?

PR #141, opened 2016-08-25
Make  SendAudio, SendDocument, SendPhoto, SendSticker, SendVideo, Sen…

kuptservol:
…dVoice to extend BotApiMethod<Message>


rubenlagus:
Did u tested it if you try to send a new file?

kuptservol:
Actually, i'm using library as the source of latest java telegram api domain objects, not invoking it through framework and just detected that some send message api objects are not built in general way.

rubenlagus:
Indeed, I didn't built them as botapimethod because they can't be sent as a pure json payload. Once this is supported by Telegram (if it is supported as some point), I'll change it.

PR #145, opened 2016-09-21
Added docs folder for new explanations

rkr8:
Perhaps it would be a good idea, to move all the explanations (e.g. HOWTO.md) into a new (docs) folder.


rkr8:
A complete Java documentation, that can be used in the most popular IDEs, would be cool too.

rubenlagus:
Tend to agree that this kind of docs may be better in the wiki than in the code. I'll see when next version goes live.

rubenlagus:
I've been thinking and indeed, every tutorial and docs must be moved to the Wiki. I'm closing this pull request. 

Anything you think should be added to the Wiki, just shout in our Telegram group for now

Issue #162, opened 2016-11-17
Missing mandatory field in inline answers.


limitium:
Hi, There is an error with serializer. When I try to answer inline mandatory field "type" is missing for InlineQueryResult objects

```java
InputTextMessageContent inputTextMessageContent = new InputTextMessageContent();
inputTextMessageContent.setMessageText("cvv");

InlineQueryResultArticle inlineQueryResultArticle = new InlineQueryResultArticle()
        .setTitle("qwe")
        .setId("1")
        .setDescription("wer")
        .setInputMessageContent(inputTextMessageContent);

AnswerInlineQuery answerInlineQuery = new AnswerInlineQuery();
answerInlineQuery
        .setResults(inlineQueryResultArticle)
        .setInlineQueryId("2");
answerInlineQuery(answerInlineQuery);
```

```json
{
  "inline_query_id": "2",
  "results": [
    {
      "@class": "org.telegram.telegrambots.api.objects.inlinequery.result.InlineQueryResultArticle",
      "id": "1",
      "title": "qwe",
      "input_message_content": {
        "@class": "org.telegram.telegrambots.api.objects.inlinequery.inputmessagecontent.InputTextMessageContent",
        "message_text": "cvv"
      },
      "description": "wer"
    }
  ],
  "method": "answerInlineQuery"
}
```

PR #163, opened 2016-11-17
Fix InlineQueryResults serialization

limitium:
Fix serialization  for https://github.com/rubenlagus/TelegramBots/issues/162


limitium:
Jackson doesn't serialize static fields

PR #177, opened 2016-12-18
Allow insecure HTTP webhook server

Tillerino:
This change allows passing a `null` keystore for the webhook server. This will start an insecure HTTP server instead of an HTTPS server. It is common to handle HTTPS encryption through a proxy, which makes forcing this onto the bot itself very awkward.


rubenlagus:
Will merge this, but using a different TelegramBotsApi constructor instead of allowing nulls

PR #187, opened 2017-01-11
fixed EditMessageText deserialization 

flicus:
when telegram returns boolean instead of edited message in response


rubenlagus:
This will be implemented changing those method to `BotApiMethod<Serializable>`. Message object shouldn't be responsible to know if a method succeeded or not.

PR #192, opened 2017-02-05
Added MonsterDeveloper's book and shield

MonsterDeveloper:
Add book about writing Java Telegram Bots with this library and added additional book shield


rubenlagus:
Merged this against dev branch. Added it to the wiki page to keep all docs together

PR #209, opened 2017-03-23
ExecutorService decoupled

saian-nimaev:
Sometimes you need to customize executorService a little. Current impl is just hardcoded. And created in constructor executorService was not being released when closing a bot session.
Also possibly `HandlerThread` and `ReaderThread` should be refactored to runnables and be submitted to separate executorService.


rubenlagus:
I'll have a look to this after next version, thx

PR #211, opened 2017-03-24
output is only 6MB

smaznet:
Now the output is only 6 MB
removed `fcom.fasterxml.jackson` and ‍`guice` will by replaced ASP :))

**NOTE** : guice used 90% of output size 

screenshot from 2017-03-25 02-17-16


smaznet:
i forgot to say it.... but webhook removed only for usage of space :////
image

because i think no needed to it ....the .jar is always running what id difference from WebHook or LongPolling?

4sef:
fucking amazing bruh @smaznet

MonsterDeveloper:
"Amazing"! Let's remove webhooks!

rubenlagus:
You made a push request with your own test bot included and against Master branch. No way this will be merged like that.

rubenlagus:
Not merging, it remove functionality, add a lot of content that should not be part of the library, remove part of pom file and is made against master branch.

I'll study in the future the posibility of replace glassfish by a different/smaller framework.

Issue #232, opened 2017-05-13
No getLevel method for BotLogger 


liketechnik:
As stated above, the BotLogger class does not have a method for retrieving the log level, although it has a method for setting it. My question is if this is intended or it just was forgotten?

PR #233, opened 2017-05-13
Add getLevel Method to BotLogger

liketechnik:
Added a getLevel method to BotLogger (see rubenlagus/TelegramBots#232).


rubenlagus:
Merged via #244

PR #248, opened 2017-06-06
Changed the constructor, so that subclasses get a chance to fully initialize themselves

asciimo71:
The super constructor creates a CommandRepo and the command repo receives the bot name from the abstract implementation. 
This calls a non-initialized method on the implementing class before the constructor returns. 
I have added a constructor with an explicit name parameter to avoid the call.


rubenlagus:
I'm not accepting pull requests against **master** branch, do it again **dev**

xMandrake:
fix this please, bots commands don't work in groups
**/command@botName**

asciimo71:
As xMandrake already pushed the change to dev, I am closing this one.

PR #249, opened 2017-06-09
Added constructor, to be able to set botname in subclasses

xMandrake:
new constructor in TelegramLongPollingCommandBot, to be able to set a botname in subclasses

### DIFF
--- a/telegrambots/src/main/java/org/telegram/telegrambots/bots/TelegramLongPollingCommandBot.java
+++ b/telegrambots/src/main/java/org/telegram/telegrambots/bots/TelegramLongPollingCommandBot.java
@@ -44,10 +44,7 @@ public abstract class TelegramLongPollingCommandBot extends TelegramLongPollingB
      * @param options Bot options
      * @param allowCommandsWithUsername true to allow commands with parameters (default),
      *                                  false otherwise
-     *
-     * @deprecated When calling this constructor, method getUsername() is not yet initialized in a subclass
      */
-    @Deprecated
     public TelegramLongPollingCommandBot(DefaultBotOptions options, boolean allowCommandsWithUsername) {
         super(options);
         this.commandRegistry = new CommandRegistry(allowCommandsWithUsername, getBotUsername());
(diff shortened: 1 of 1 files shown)


rubenlagus:
@xMandrake Why isn't it enough to override getUsername method?

xMandrake:
Because overrided method getName() will be called after Subclass constructor, so getName() in absrtract class will be null
Simple example:
https://gist.github.com/xMandrake/cbf05577e5884fe74fcfd2f1b0521f87#file-example-java

asciimo71:
Thanks xMandrake to bring the fix to dev. I tried to but I am new to github and my git knowledge is rusty...

rubenlagus:
After checking this, I don't like the idea of having the possibility of using one username for the constructor and a different one in getUsername method. Both of them should be the same. 

The only fast solution I can imagine is implementing `getUsername` method with final modifiers and deprecating all constructors without username.

Other option is to change the CommandRegistry so it accept a function instead of a String as username, but this one may require more changes.

What do you think?

asciimo71:
I agree, my change is not perfect - I have also seen exactly the problem 
you describe. I left it to the developer to keep the name in sync.

To fix this properly, I don't see, why the name should be implemented as 
a getter in the implementing class if it is not allowed to be changed 
later.  The getter approach suggests, that you could calculate the 
username from some magic, but in the end, you can't. Because, if the 
magic relies on the constructor to have happened, the magic will fail. 
The same applies to the lambda approach: It suggests that you can 
influence it later.

The username is final - you do not allow setting it after the 
constructor has run - so the interface should reflect this.

Having said and suffered this, I strongly support your idea to remove 
all constructors without a username and have the developer do the 
username magic before creating the object. This reflects much more the 
way things actually work. The getUsername() function needs not to be 
abstract any more, as you said, protecting it with a final modifier is a 
clear signal in the interface as well.

Thumbs up!

Best regards,

Thomas

Am 18.06.2017 um 04:49 schrieb Ruben Bermudez:

PR #255, opened 2017-06-21
Fix API error on UTF-8 encoded performers / titles

devbridie:
Sending an UTF-8 encoded string as an Audio's title or performer would result in the following error: `org.telegram.telegrambots.exceptions.TelegramApiRequestException: Error sending audio: [400] Bad Request: audio title must be encoded in UTF-8`. 

This is because the default `ContentType` in `addTextBody` is a `ContentType.DEFAULT_TEXT`, which has Charset `ISO_8859_1`.

This fix sets the type to `TEXT_PLAIN_CONTENT_TYPE`.


devbridie:
Accidentally came from `master` instead of `dev`. See #256 instead.

PR #256, opened 2017-06-21
Fix API error on UTF-8 encoded performers / titles

devbridie:
Sending an UTF-8 encoded string as an Audio's title or performer would result in the following error: `org.telegram.telegrambots.exceptions.TelegramApiRequestException: Error sending audio: [400] Bad Request: audio title must be encoded in UTF-8`. 

This is because the default `ContentType` in `addTextBody` is a `ContentType.DEFAULT_TEXT`, which has Charset `ISO_8859_1`.

This fix sets the type to `TEXT_PLAIN_CONTENT_TYPE`.

### DIFF
--- a/telegrambots/src/main/java/org/telegram/telegrambots/bots/DefaultAbsSender.java
+++ b/telegrambots/src/main/java/org/telegram/telegrambots/bots/DefaultAbsSender.java
@@ -565,10 +565,10 @@ public abstract class DefaultAbsSender extends AbsSender {
                     builder.addTextBody(SendAudio.REPLYTOMESSAGEID_FIELD, sendAudio.getReplyToMessageId().toString());
                 }
                 if (sendAudio.getPerformer() != null) {
-                    builder.addTextBody(SendAudio.PERFOMER_FIELD, sendAudio.getPerformer());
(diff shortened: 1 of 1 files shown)


rubenlagus:
Nice @devbridie, maybe it makes also sense to add this fix to all the Caption fields?

devbridie:
@rubenlagus I have added the fix the `caption` fields for Audio and Voice. Aside from these I have not found any other cases that display the behavior. 

Concerning #257, it seems lines [196](https://github.com/devbridie/TelegramBots/blob/95570da559e48eaf9e7136143a6927cf1b35c612/telegrambots/src/main/java/org/telegram/telegrambots/bots/DefaultAbsSender.java#L196), [198](https://github.com/devbridie/TelegramBots/blob/95570da559e48eaf9e7136143a6927cf1b35c612/telegrambots/src/main/java/org/telegram/telegrambots/bots/DefaultAbsSender.java#L198), and [200](https://github.com/devbridie/TelegramBots/blob/95570da559e48eaf9e7136143a6927cf1b35c612/telegrambots/src/main/java/org/telegram/telegrambots/bots/DefaultAbsSender.java#200) are related. However, `addBinaryBody` does not seem to have the capability to set the encoding for the `filename` field.

PR #272, opened 2017-07-20
Refactor default abs sender

fedochet:
I was studying your library and spotted some code duplication in `DefaultAbsSender`, so I decided to refactor it
I did not  change any behavior - only rearranged some lines in order to make refactorings clearer


fedochet:
I am not sure why CodeCov fails - considering I didn't change mentioned files :smile:

PR #284, opened 2017-08-02
#269 Fixed update queue polling order and added `TelegramBatchLongPollingBot`

vantuz-subhuman:
1. Added method to `LongPollingBot` that accepts a list of updates and by default calls standard `onUpdate` method sequentially.
2. `DefaultBotSession` is changed to poll the whole update queue and feed it into the new method. The polling method is safe for multithreading.

@rubenlagus


vantuz-subhuman:
@rubenlagus, I had to refactor some code inside `DefultBotSession` in order to make it testable. But it allowed me to create two test methods that cover the functionality of storing updates and sending them to long polling bots (both default and batch one).

I don't know how to recalc codecov report =(

rubenlagus:
@vantuz-subhuman Don't worry

rubenlagus:
@vantuz-subhuman I was wondering. Why add the new batchpollling class? Why don't use the default longpollingbot with a default implamentation that just delegate the updates from *onUpdatesReceived(List<Update>)* to the *onUpdateReceived(Update)* one by one?


At the end, anyone that needs the batch version, just need to override *onUpdatesReceived(List<Update>)* method.

This way, it will simplify the BotSession since it will only have one single way of fetching the updates: It will always delegate the full list to the bot and the bot should sort out the order.

What do you think?

vantuz-subhuman:
@rubenlagus, I totally agree, it's just that I tried minimally affect existing classes, and kinda just add new stuff as side thing. But if you're up for a change like this - I will happily change this PR.

rubenlagus:
I prefer to keep only one class for this, in addition, the actual impact in existing bots should be none since we still keep current behavior as default one

vantuz-subhuman:
@rubenlagus, I've finally been able to add the changes we discussed. Now there's no additional specific bot type, only new method in the basic `LongPollingBot`

vantuz-subhuman:
Closed by mistake. Missed the button (

Issue #304, opened 2017-09-14
CommandRegistry formats BotCommand parameters incorrectly


dschulz:
Hi,

I just noticed that a command like

`/command@bot___param1_param2______param3`

or simply

`/command___param1_param2______param3`

results in a `String [] parameters` with spurious empty `String`s  getting passed to the `execute()` method. Assuming the above example (with _ characters replaced by actual spaces) the result is

```java
System.out.println("Parameters length: " + parameters.length 
+ "\nParameters: [" + String.join("|", parameters) + "]");
```
> Parameters length: 10
> Parameters: [||param1|param2||||||param3]

The correct parameters should be

> Parameters length: 3
> Parameters: [param1|param2|param3]

This causes `BotCommand` objects to receive wrong parameters when users send commands that expect arguments.

dschulz:
My current workaround for this problem is to add the following in every `BotCommand` subclass:

```java
@Override
public void execute(AbsSender absSender, User user, Chat chat, String[] origArgs) {
   String[] args = Arrays.asList(origArgs).stream().filter(e -> !(e.trim().isEmpty())).collect(Collectors.toList())
				.toArray(new String[0]);

   // ...
}

```

which is clearly ugly 😁

Fotomultman:
Nice try)

rubenlagus:
Should be already fixed

PR #309, opened 2017-09-18
Use a regular expression to split command parameters

dschulz:
Use a regular expression instead of a single space to split command parameters. This avoids spurious/empty parameters getting passed to the BotCommand `execute` method.


dschulz:
Proposed fix for #304

PR #317, opened 2017-10-16
SendDocument from external URL

tom139:
This commit will add the ability to send a document from an external URL source, as allowed by Telegram API (see [sendDocument method](https://core.telegram.org/bots/api#senddocument)).

The first change I did is in the `SendDocument` class. I spent some time thinking whether the right approach would have been to allow using the existing method for sending older files by file_id `public SendDocument setDocument(String document) {}` or to add a new method. I went for the latter with:
```java
public SendDocument setURL(String url) {
    	this.document = url;
    	this.isNewDocument = true;
    	return this;
    }
```
I think this is more consistent with the way the API is structured.

The next step was to extend the _if_ in `SendDocument.validate()` and `DefaultAbsSender.sendDocument`.

Please let me know what you think about this change and how I can improve this pull request (if this will not be merged).

Thank you for this great library


rubenlagus:
@tom139 Wasn't `setDocument(String)` already valid for an URL?

tom139:
I'm sorry, I tried with another document url and it works. There might be a problem with the Telegram API but the following code is working almost every time:
```cURL
curl -X POST \
  https://api.telegram.org/bot<token>/senddocument \
  -H 'content-type: application/x-www-form-urlencoded' \
  -d 'chat_id=<chat_id>&document=https%3A%2F%2Finstagram.fmxp1-1.fna.fbcdn.net%2Ft51.2885-15%2Fe35%2F22344305_128171461238189_1320416175359787008_n.jpg'
```
But the following code works almost never 
```java
SendDocument d = new SendDocument();
d.setChatId(chatID);
d.setDocument("https://instagram.fmxp1-1.fna.fbcdn.net/t51.2885-15/e35/22344305_128171461238189_1320416175359787008_n.jpg");
```
I get `org.telegram.telegrambots.exceptions.TelegramApiRequestException: Error sending document: [400] Bad Request: wrong type of the web page content`.

I'm sorry but the change I proposed is probably unuseful.

rubenlagus:
Maybe some characters escaping must be done?

tom139:
The problem lies outside the scope of the bots API, so will close this pull request.
Thank you again!

PR #328, opened 2017-11-01
Implement DefaultBotCommand, which permit to process user message dir…

vadimgoroshevsky:
…ectly in "execute" method


rubenlagus:
From my point of view, it makes more sense that the base class of a command is called `BotCommand` and not `DefaultBotCommand`. The former sounds more like a default implementation. What do you think @Mit0x2 ? (Since you were the first implementer of command bot)

addo37:
I concur with Ruben on this. DefaultBotCommand should be the one adding small "increments" to its parent class. BotCommand is just base.

vadimgoroshevsky:
Thank you for review guys. I will open new PR later.

PR #340, opened 2017-11-19
#325 Implement equals() and hashCode() in SendMessage

### LINKED ISSUE #325: Why SendMessage does not implement equals() and hashCode()?
I was writing tests for the bot, checking the output and found out that there is no equals() and hashCode() method. So I would need to use refEq from Mockito framework to pass the test. Is there a particular reason not to implement equals() and hashCode() methods?

myplacedk:
Closes #325

### DIFF
--- a/telegrambots-meta/src/main/java/org/telegram/telegrambots/api/methods/send/SendMessage.java
+++ b/telegrambots-meta/src/main/java/org/telegram/telegrambots/api/methods/send/SendMessage.java
@@ -188,6 +188,35 @@ public class SendMessage extends BotApiMethod<Message> {
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


rubenlagus:
@myplacedk shouldn't your solution also include implementations for the different possible reply_markups?

myplacedk:
Ah yes, I forgot to check if it depended on other classes, sorry.

myplacedk:
Codecov seems broken. The code it says is no longer covered doesn't seem related to this PR at all.

I didn't make any unit tests specifically for my added code, I don't feel it would add enough value for the time it takes to create and maintain them. Let me know if you don't agree.

PR #391, opened 2018-02-23
Update Using-Replies.md

Baspla:
The Consumer has the variable name of "action" not "upd".


addo37:
Thank you for the PR. 👍

Issue #429, opened 2018-04-05
How to set it up to connect using SOCKS5 proxy?


bvn13:
Is there any way to do that?

zugzug90:
@bvn13 Maybe we can customize something in https://github.com/rubenlagus/TelegramBots/blob/master/telegrambots/src/main/java/org/telegram/telegrambots/bots/DefaultBotOptions.java? It uses org.apache.http.client.config.RequestConfig in its guts.. that's the starting point, I guess, we can dig into. But not sure about SOCKS5.. maybe just HTTP(S) can be applied somehow.

    /**
     * @implSpec Default implementation assumes no proxy is needed and sets a 75secs timoute
     * @param requestConfig Request config to be used in all Http requests
     */
    public void setRequestConfig(RequestConfig requestConfig) {
        this.requestConfig = requestConfig;
    }

ospavel:
Is there ability to change socks5/proxy on the fly without recreating bot?
For example, if working bot catches "Connection timed out" then just set new socks5/proxy.

rubenlagus:
@ospavel There is no support for socks5 yet, just for proxies. But not, at the moment, there is no way to change proxy without recreating the bot.

carlopantaleo:
@ospavel @rubenlagus There's a workaround to set a SOCKS5 proxy, see [here](https://github.com/rubenlagus/TelegramBots/pull/435#issuecomment-382722476). It works, as I'm currently using it. However, in this way I think it is set at JVM level, so all requests (except those to localhost) are proxied.

rubenlagus:
@bvn13 This should be solved already, can be closed?

bvn13:
@rubenlagus I'll check it

bvn13:
@rubenlagus 

I'm trying to use 4.0.0 and 4.0.1 versions now and don't see any way to set uputhentication for proxy.

Is this ability broken?

PR #451, opened 2018-05-06
Socks proxy feature

AzZureman:
I deleted apache.http proxy because apache.http doesn't support proxy (cause http and socks work on different levels), and added java.net proxy throug the ConnectionManager. So now it's also support  socks proxy.
I also update proxy usage page on the wiki and change bots version. (cause new version does not support the old methods of enabling a proxy).
The working ability of all types of proxy (http with/without auth and socks with/without auth) was manually tested by me.


AzZureman:
This pull-request closes issue #429

rubenlagus:
@AzZureman I've been checking this changes. It is breaking backward compatibility completely, do you think is there any option to prevent that? I'd like to keep versions together with Official Telegram Bots API, and adding this kind of changes will force and update to Version 4.0.

Azim-D3Tm:
@AzZureman does it support both SOCKS and http proxies or only SOCKS?

PR #454, opened 2018-05-09
Add detailed constructor, equals and hashCode to User object

addo37:


rubenlagus:
@addo37 The fields in User object (as any other object coming from Telegram) is not supposed to be modified by the library, so they don't contains any setter or constructor with parameters. Those object will just be build when parsing the JSON from Telegram servers

On the other hand, I'm ok with equals and hash methods.

addo37:
@rubenlagus I totally understand that. However, the main reason for introducing a detailed constructor is to give the user the ability to create fake users without mocking the class itself. It's out of convenience, nothing more nothing less.

Also, I suggest adding the final modifier on the fields as well as a @JsonCreator for a constructor/static factory method. We require immutability and this is one way of ensuring it. Jackson actually uses reflection to set those private fields when deserializing.

carlopantaleo:
> [...] the main reason for introducing a detailed constructor is to give the user the ability to create fake users without mocking the class itself.

In this way you are introducing test logic in main code, just for convenience for testing. Isn't it wrong?

carlopantaleo:
@addo37 This discussion is getting interesting.
I inspected the code more deeply and I agree with you indeed. If ``User`` is a POJO, it should agree to the JavaBean specification. Since it's supposed to be immutable, it's ok it only has getters, but it's better if it has a method to construct it without accessing private fields with reflection.

Having a parametrized constructor doesn't break the requirement of immutability. It's true that the library could produce an ad-hoc object which doesn't reflect the original JSON, but it could have done it anyways even with a default constructor by using reflection, as Jackson does.

PR #460, opened 2018-05-17
fix field name pending_update_count of WebhookInfo

azhohov:
I corrected field name pending_update_count of WebhookInfo.
https://core.telegram.org/bots/api#getwebhookinfo
pending_update**s**_count -> pending_update_count


rubenlagus:
Nice catch @azhohov  👍

PR #487, opened 2018-06-25
Added ability to provide and use a SOCKS5 proxy only for TelegramBots

vogoltsov:
The problem with using telegram proxies is most of them are SOCKS5-based which renders current solution almost unusable: you either have to provide a JVM-level http proxy, or use transparent http proxies (which is not a very good solution).

This implementation is an example how you can force TelegramBots to use a SOCKS5 proxy using existing configuration facilities (https://github.com/rubenlagus/TelegramBots/wiki/Using-Http-Proxy). The only that is different from HTTP proxy configuration is you have to define "socks5" as a scheme for a proxy host like this:
```java
            defaultBotOptions.setHttpProxy(
                    new HttpHost(
                            "...",
                            "...",
                            "socks5"
                    )
            );
```

Honestly, I think the ```TelegramHttpClientBuilder``` should be heavily refactored as it cannot be extended or customized through the API. It is also not possible to provide a custom http client builder to ```DefaultAbsSender``` and ```DefaultBotSession```.

### DIFF
--- a/telegrambots/src/main/java/org/telegram/telegrambots/facilities/ProxyPlainConnectionSocketFactory.java
+++ b/telegrambots/src/main/java/org/telegram/telegrambots/facilities/ProxyPlainConnectionSocketFactory.java
@@ -1,0 +1,30 @@
+package org.telegram.telegrambots.facilities;
+
+import org.apache.http.conn.socket.PlainConnectionSocketFactory;
+import org.apache.http.protocol.HttpContext;
+
+import java.net.Proxy;
+import java.net.Socket;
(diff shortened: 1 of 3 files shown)


Clevero:
The [README](https://github.com/rubenlagus/TelegramBots#contributions) states:

```Feel free to fork this project, work on it and then make a pull request against DEV branch.```

This PR is against master xD

vogoltsov:
Recreated PR against dev: https://github.com/rubenlagus/TelegramBots/pull/488

Issue #480, opened 2018-06-18
What about Java9+?


bvn13:
Have an error:
module 'my-module' reads package 'org.telegram.telegrambots' from both 'telegrambots' and 'telegrambots.meta'.

Seems it need to reneme sub-projects' packages like org.telegram.telegrambotsmeta or place it into one project (worse).

rubenlagus:
Java9 is, according to Oracle a `short term release`, quite probably I'll update to Java10 directly at some point in the future. Not planned at the moment.

ChmilevFA:
I have the same issue for the Java 10

ChmilevFA:
Do we have any ways to use with Java 9+?

ChmilevFA:
A bit more details:
- trying to use Java 9 and Java 10 (and I suppose Java 11, because it has no changes as far as I understand) with standard module feature using module-info.java
- trying to use classes from telegrambots and telegrambots-meta modules and "import" them into module-info.java
```
module your.module  {
    ...
    requires telegrambots.meta;
    requires telegrambots;
    ...
}
```
You will get this error:

> module 'your.module' reads package 'org.telegram.telegrambots' from both 'telegrambots' and 'telegrambots.meta'.

Seems like it is impossible to use the same package name from different modules with Java's module system.

Besides the above mentioned problem with using the same package name for different modules seems bad practice just because you must know which classes you have in another modules not to intersect names for them.

rubenlagus:
@bvn13 @ChmilevFA Can we close this issue with last release?

bvn13:
@rubenlagus not checked yet

PR #491, opened 2018-07-08
Java9+ support

ChmilevFA:
It will fix errors described in #480 

I just moved:
- all from `org.telegram.telegrambots` package in **extensions** module into `org.telegram.telegrambots.extensions`
- all from `org.telegram.telegrambots` package in **meta** module into `org.telegram.telegrambots.meta`

It will break backward compatibility but I think it's necessary to work with module system of new versions of Java.
With my suggestion for this PR clients just have to edit some imports.

Another way to solve this problem is to move all the same packages into the same module.


rubenlagus:
@ChmilevFA Can you update the PR to fix conflicts? Quite probably merging for next release

ChmilevFA:
@rubenlagus done
Also added commit for proper update version to 3.6.2 because https://github.com/rubenlagus/TelegramBots/commit/ff73a433375ef98b235e1a527ee3d3b88d296c4d#diff-600376dffeb79835ede4a0b285078036 seems as incomplete update

Could you please merge it ASAP (or before others PR in dev)? Because it can be harmful solve all conflicts quick and often for this changes.

rubenlagus:
@ChmilevFA I'll check and merge this one before any other

Issue #536, opened 2018-11-04
execute(SendPhoto) fails when bot is injected as Spring dependency


donkeystep:
Could you please make DefaultAbsSender.execute() methods not final?
(When bot is an injected dependency, AOP can't call them.)

The difference is:
* AbsSender.execute(Method) - not final, works ok 
![selection_074](https://user-images.githubusercontent.com/20819782/47968708-aa273f00-e07e-11e8-8a59-b7d20ccd94d0.png)
* DefaultAbsSender.execute(SendPhoto) - final, AOP proxy cannot call the real class
![selection_075](https://user-images.githubusercontent.com/20819782/47968719-c6c37700-e07e-11e8-83eb-b930ddf56755.png)
```
java.lang.NullPointerException: null
	at org.telegram.telegrambots.bots.DefaultAbsSender.sendHttpPostRequest(DefaultAbsSender.java:765) ~[telegrambots-4.1.jar:na]
	at org.telegram.telegrambots.bots.DefaultAbsSender.execute(DefaultAbsSender.java:226) ~[telegrambots-4.1.jar:na]
	at org.vors.telegram.fivemin.service.MessageService.sendPhoto(MessageService.java:107) ~[classes/:na]
```
https://stackoverflow.com/questions/34563058/spring-aop-cglib-proxys-field-is-null

# As a workaround, I added such method to my bot class:
```
public Message sendPhoto(SendPhoto sendPhoto) throws TelegramApiException {
    return execute(sendPhoto);
}
```

PR #537, opened 2018-11-05
#536 fix "execute(SendPhoto) fails when injected as Spring dependency"

donkeystep:
Fix for #536 - removed final from public methods to let Spring AOP proxy call the implementation


Clevero:
Please change the pull request against DEV branch.

The README states: ```Feel free to fork this project, work on it and then make a pull request against DEV branch. ```
Thanks!

rubenlagus:
Those methods are set as final on purpose since they should not change or being override. They contains the API layer of the implementation and allowing overriding it can cause undesired behaviour if someone does it by mistake.

donkeystep:
Thanks @rubenlagus , my fix is wrong. Maybe it makes sense to extract the methods to interface to let Spring use Dynamic Proxy.

There's workaround that eliminates the issue for all calls except downloadFile* - use AbsSender type for dependency:
```
@Autowired
@Lazy
private AbsSender bot;
```

PR #541, opened 2018-11-06
Fix typos in comments, error messages and private fields

aNNiMON:
These changes are safe and won't affect the end user (unless somebody uses exception messages in tests). However, I found a set of typos in getters/setters and class names:

 - https://github.com/rubenlagus/TelegramBots/blob/bdc8db3c87cdbad2f3e1824f64d3496066049908/telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/methods/updatingmessages/EditMessageLiveLocation.java#L116-L118

 - https://github.com/rubenlagus/TelegramBots/blob/bdc8db3c87cdbad2f3e1824f64d3496066049908/telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/methods/groupadministration/DeleteStickerSetName.java#L23-L24 

 - https://github.com/rubenlagus/TelegramBots/blob/bdc8db3c87cdbad2f3e1824f64d3496066049908/telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/methods/groupadministration/GetChatMemberCount.java#L20-L21

There are two ways to fix these issues. One is to create a copy of class/method and mark an old copy as deprecated for removal. This will keep compatibility for the next version. Second is to break compatibility and simply rename these methods and classes. What do you prefer?

Do I need to create another pull request for listed changes?

### DIFF
--- a/telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/TelegramBotsApi.java
+++ b/telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/TelegramBotsApi.java
@@ -18,7 +18,7 @@ public class TelegramBotsApi {
     private static final String webhookUrlFormat = "{0}callback/";
     private boolean useWebhook; ///< True to enable webhook usage
     private Webhook webhook; ///< Webhook instance
-    private String extrenalUrl; ///< External url of the bots
+    private String externalUrl; ///< External url of the bots
     private String pathToCertificate; ///< Path to public key certificate
 
     /**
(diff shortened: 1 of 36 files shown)


rubenlagus:
I'd prefer to create a copy and mark them as deprecated. We can remove them in next major version. 

Feel free to use this same pull request for them

aNNiMON:
I have noticed one more typo, now in package name. I've added a fix to this branch

PR #543, opened 2018-11-07
Enable AbilityBot to accept methods which return List<Ability> and add these abilities.

skiclub-developer:
Resubmitted Pull request against dev branch.

For my bot I want to fetch a number of commands from my database and return methods for these commands dynamically. Thus I needed the AbilityBot to be able to accept methods which return List.

Would be nice to have this functionality added.


addo37:
Hey skiclub-developer!

Thanks for submitting a PR for the ability bot module. While I do see the benefit of adding this feature, I think we'll need to work a bit further to have it integrated into the codebase.
1 - I'd rather support an Iterable over a List
2 - Feature must be tested - take a look at the available tests in the framework

Besides that, there are a few technical issues with the abilities construction. Relying on a null return value to verify class cast is not conventional. Moreover, abilities map should not be built twice. If I do find the time, I'll be submitting a PR on your dev branch.

addo37:
@skiclub-developer Can you please rebase and push?

PR #599, opened 2019-03-29
Back up maps in the database as a collection of key value pairs

bernikr:
Because the JSON (de)serializer only uses strings as keys, there are problems when using other types as keys. (As indicated by the workaround for the USERS map with long keys in the old code)

By backing up all maps as a collection of key value pairs and restoring them correctly as maps, all serializeable objects could be used as keys without problems.

### DIFF
--- a/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/db/BackupMap.java
+++ b/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/db/BackupMap.java
@@ -1,0 +1,36 @@
+package org.telegram.abilitybots.api.db;
+
+import org.telegram.abilitybots.api.util.Pair;
+
+import java.util.*;
+
+public class BackupMap<K, V> extends AbstractCollection<Pair<K, V>> implements Collection<Pair<K, V>> {
+    private Collection<Pair<K, V>> entries = new HashSet<>();
+
+    public BackupMap(){}
+
+    public BackupMap(Map<K, V> map) {
+        map.forEach((key, value) -> entries.add(Pair.of(key, value)));
+    }
+
+    public Map<K, V> toMap() {
+        Map<K, V> map = new HashMap<>();
+        entries.forEach(e -> map.put(e.a(), e.b()));
+        return map;
+    }
+
+    @Override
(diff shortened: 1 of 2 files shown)


telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/db/BackupMap.java
```diff
@@ -0,0 +1,36 @@
+package org.telegram.abilitybots.api.db;
+
+import org.telegram.abilitybots.api.util.Pair;
+
+import java.util.*;
+
+public class BackupMap<K, V> extends AbstractCollection<Pair<K, V>> implements Collection<Pair<K, V>> {
```
addo37: Make class final, I don't see a use case beyond recovery

telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/db/MapDBContext.java
```diff
@@ -197,17 +197,8 @@ private void doRecover(Map<String, Object> backupData) {
...
-        // TODO: This is ugly
-        // Special handling of USERS since the key is an integer. JSON by default considers a map a JSONObject.
-        // Keys are serialized and deserialized as String
-        if (name.equals(USERS))
-          entryMap = entryMap.entrySet().stream()
-              .map(entry -> Pair.of(Integer.parseInt(entry.getKey().toString()), entry.getValue()))
-              .collect(toMap(Pair::a, Pair::b));
-
+      } else if (value instanceof BackupMap) {
+        Map<Object, Object> entryMap = ((BackupMap)value).toMap();
```
addo37: Add space in between `(BackupMap) and value `

telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/db/BackupMap.java
addo37: This is supposed to be an internal class used for backup and recovery; class should be package private.

addo37:
Thanks @bernikr, this should've been done waaaay earlier. 👍

PR #601, opened 2019-04-10
Extra fields appear during serialization meta objects

galimru:
I want to get the raw update from telegram api using your library.
The only possible way I found is to serialize Update DTO to JSON.
However, when I did it in JSON content appeared extra fields like `groupChat`, `userChat` and so on.
Also, for User's meta the bot field appears twice as `bot` and as `is_bot`.

I think custom fields should be ignored for serialization.


rubenlagus:
I can't merge it directly, but will do it for next release

rubenlagus:
@rgalimov Can you please undo the changes to the pom files? The rest looks fine. If so, I´ll just merge it directly.

galimru:
I will prepare another PR

PR #604, opened 2019-04-14
Added new fields

Systemoutprintln7:
Added new fields of 4.2 API update for ChatMember and Message objects.


rubenlagus:
Were added as part of the full update

PR #631, opened 2019-06-27
Fixed case insensitive usernames

jugendhacker:
Usernames in Telegram are case insensitive, eg fooBar is the same as FooBar.
So this bot framework should also ignore the case of the username and this PR fixes this.


jugendhacker:
Huh why did Travis fail with this strange error...?

rubenlagus:
@jugendhacker This change is a breaking change, please change it so it deprecates current methods before removing any method.

jugendhacker:
@rubenlagus You mean the constructors that are breaking?

rubenlagus:
yes

jugendhacker:
I now added an constructor with the old style, but marked as deprecated

PR #633, opened 2019-07-09
AbilityBots: prevent nullPointerExceptions when using message flags

bernikr:
When using only a message flag like `LOCATION` without `MESSAGE` first, a nullPointerException is thrown if a non-message update is received.

I believe that instead of throwing the exception, the massage flags like `LOCATION` or `REPLY` should just return false. This prevents bot developers from having to specify the redundant `MESSAGE` flag.

This PR fixes this bug by checking `hasMessage()` before calling `getMethod()`.


addo37:
Hey @bernikr!

Can you please modify all the message flags as such:
Reply modification, from:
`REPLY(upd -> upd.hasMessage() && upd.getMessage().isReply())`
To:
`REPLY(upd -> MESSAGE.test(upd) && upd.getMessage().isReply())`

I'm totally fine with this change. I can perform a PR with those changes to your dev branch if you'd like to.

PR #660, opened 2019-09-09
Replacing log4j with slf4j

UnAfraid:
Using slf4j abstraction instead of log4j directly
slf4j is just an abstraction layer, it allows you to use number of logging implementations including log4j2.
Which leaves the developers to choose which logging library suits them better.

Configuration remains the same, existing projects may just need to change their imports and LogManager -> LoggerFactory and Logger#fatal -> Logger#error


TopoDiFogna:
This should be an improvement to the library since it allows devs to choose their favourite logging framework

PR #664, opened 2019-09-17
Support backup and recovery of db vars

### LINKED ISSUE #642: DBContext recover with Var fails
Using db.getVar("...") and trying to create a backup, this saves it as ["org.mapdb.Atomic$Var",{"recid":42}] and since there is not really any data to recver anything the recovery fails.

_Part of the exception stack:_
InvalidDefinitionException: Cannot construct instance of `org.mapdb.Atomic$Var` (no Creators, like default construct, exist): cannot deserialize from Object value (no delegate- or property-based Creator)

addo37:


addo37:
Closes #642

PR #676, opened 2019-10-03
Create system.properties

usalavat:


rubenlagus:
What's the purpose of adding this file? Maven should already use java 11

PR #682, opened 2019-10-14
Fixed maven and gradle imports

abramchikd:
There is no 4.4.0 version available only 4.4.0.1
https://stackoverflow.com/questions/57669668/could-not-resolve-org-telegramtelegrambots4-4-0


rubenlagus:
This is against master

Issue #655, opened 2019-08-27
The artifact `org.telegram:Bots:4.4.0` is missing in maven repo


peterquiel:
Motivated by https://stackoverflow.com/questions/57669668/could-not-resolve-org-telegramtelegrambots4-4-0/57670630#57670630

If you check out https://mvnrepository.com/artifact/org.telegram/Bots you will notice that `4.4.0` is missing.

rubenlagus:
Use version 4.4.0.1, it should be fixed there

PR #686, opened 2019-10-30
Fix example considering #655

rugpanov:
Version 4.4.0 is missing in the maven central repository.
https://github.com/rubenlagus/TelegramBots/issues/655


rubenlagus:
Request against master

PR #687, opened 2019-10-30
upgrade spring boot to 2.2.1.RELEASE, replaced Optional bean with ObjectProvider in autoconfiguration

SokoMishaLov:
upgrade spring boot to 2.2.0.RELEASE
replaced Optional bean with ObjectProvider in autoconfiguration


rubenlagus:
This looks like a breaking change, please do it in a way we can deprecate old methods at least for some time

SokoMishaLov:
@rubenlagus hmm, actually i don't think so, because these bean methods in auto configuration are `spring-specific`
i doubt that someone calls it directly, most of people create their own configuration to override these default beans

PR #699, opened 2019-12-04
Fix canRestrictMembers property name in ChatMember

Djaler:
https://core.telegram.org/bots/api#chatmember


Djaler:
@rubenlagus can you check this?

rubenlagus:
Pull request against master, not merging.

I'll do the change in dev branch

Issue #716, opened 2020-01-20
TelegramLongPollingCommandBot NPE when command update is received. 


ds64:
The issue is reproduced since TelegramBotExtensions 4.4.0.2, there is a problem that botUsername field of that class is never set before creating an instance of CommandRegistry class. So, botUsername field is always null in CommandRegistry class.

When bot receives update with command in it from Telegram, CommandRegistry.removeUsernameFromCommandIfNeeded method is called. This code in CommandRegistry.java, line 108, produces NPE as botUsername is always null:

`return this.allowCommandsWithUsername ? command.replaceAll("(?i)@" + Pattern.quote(this.botUsername), "").trim() : command;`

The proposal is to either return TelegramLongPollingCommandBot constructors which receive botUsername as a string, or remove everything "@" character in command without parcing bot username itself.

Chase22:
The error is actually caused by the botoptions constructer not setting a bot username 

https://github.com/rubenlagus/TelegramBots/blob/bd5c5a2f416ed6c556e684f3829671aee1ab3cc7/telegrambots-extensions/src/main/java/org/telegram/telegrambots/extensions/bots/commandbot/TelegramLongPollingCommandBot.java#L68

It needs another parameter to pass in the bot username

ds64:
@Chase22, I am gonna check this approach. But even though I know it's deprecated, the 
`public TelegramLongPollingCommandBot(String botUsername)`
constructor is definitely broken.
First, this() is called which creates an instance of CommandRegistry under this bot instance and only after that, the botUsername field is set. So, botUsername field is never set in CommandRegistry instance even if it is passed to constructor. And any update received results in NPE.

Chase22:
@ds64 It's definetly a bug. I'll go ahead and make a fix for it. Thanks for pointing it out

Chase22:
Fixed in #735

PR #717, opened 2020-01-20
https://github.com/rubenlagus/TelegramBots/issues/716 FIX proposal

ds64:
FIX for NPE in CommandRegistry.java


Chase22:
I think the proposal would cause quite a lot of unneccessary requests to the api. The bot username never changes. You could use that either as a fallback or explicitly check if the username is set when the registry is created and interrupt the startup if it isn't

rubenlagus:
Should be fixed already with #735,

PR #730, opened 2020-03-01
Prevent loss of DB state after bot termination

addo37:
The call to `db.commit` must be done at the end of the pipeline. At the moment, it's used in the user addition logic which happens to be at the beginning of the pipeline.

Something like:
State 1 -> State 2 -> Bot Termination would make the bot start on State 1. This mostly affects ReplyFlows.


addo37:
@rubenlagus This is good to go for the next merge.

PR #734, opened 2020-03-05
Added "How to get a list of all group members" to FAQ

F1R3H4WK:


addo37:
Isn't this a bit misleading? Since you only get new chat members, it might not pick up the pre-existing members.

TelegramBots.wiki/FAQ.md
```diff
@@ -282,3 +283,24 @@ After that your bot will look like:
...
+   public class YourBotClassName extends TelegramLongPollingBot {
+     List<User> groupMembers = new ArrayList<>();
+     //Bot body.
+   }
+```
+And if someone joins the group, add a new user to the list:
+```java
+   @Override
+   public void onUpdateReceived(Update update) {
+     if (update.getMessage().getNewChatMembers() != null) {
```
Chase22: getNewChatMember always returns something. If newChatMembers is null it will return an empty list.

https://github.com/rubenlagus/TelegramBots/blob/8351707e835bea544aa183ead179bb2fda0788d2/telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/objects/Message.java#L302

rubenlagus:
In addition, this is a pull request against master, so not accepted. Do it against Dev

PR #744, opened 2020-04-06
Update wiki pages

petrov9:
Changed example's code, now it's matches the version


addo37:
@rubenlagus This is good to go for the next merge.

PR #748, opened 2020-04-24
Update version Api 4.8

rubenlagus:


telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/methods/send/SendDice.java
```diff
@@ -109,6 +121,9 @@ public void validate() throws TelegramApiValidationException {
         if (chatId == null) {
             throw new TelegramApiValidationException("ChatId parameter can't be empty", this);
         }
+        if (emoji != null && !emoji.equals("“\uD83C\uDFB2") && !emoji.equals("\uD83C\uDFAF")) {
```
kobyzau, line 124: Extra symbol near "dice" emoji - “

rubenlagus: Solved

Issue #749, opened 2020-04-25
Usage Abilities, webhook and polls causes IllegalStateException


kortov:
Create Webhook ability bot, add ability to send a poll to a user via `sendApiMethod` (reproducable with quiz and poll)

```kotlin 
val sendPoll = SendPoll(AbilityUtils.getChatId(upd), "question", Arrays.asList("a", "b", "c"))
sendApiMethod(sendPoll) 
```

when user clicks on the answer 
![image](https://user-images.githubusercontent.com/6348215/80289970-4c475500-8753-11ea-8e43-d35863f3ea06.png)

the telegram sends to a bot this

```
{
    "update_id": 689323206,
    "poll": {
        "id": "5269766460913221652",
        "question": "question",
        "options": [
            {
                "text": "a",
                "voter_count": 1
            },
            {
                "text": "b",
                "voter_count": 0
            },
            {
                "text": "c",
                "voter_count": 0
            }
        ],
        "total_voter_count": 1,
        "is_closed": false,
        "is_anonymous": true,
        "type": "regular",
        "allows_multiple_answers": false
    }
}
```

and then bot throws exception

``` java
java.lang.IllegalStateException: Could not retrieve originating user from update
	at org.telegram.abilitybots.api.util.AbilityUtils.getUser(AbilityUtils.java:73) ~[telegrambots-abilities-4.8.jar:na]
	at org.telegram.abilitybots.api.bot.BaseAbilityBot.checkBlacklist(BaseAbilityBot.java:355) ~[telegrambots-abilities-4.8.jar:na]
	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:174) ~[na:1.8.0_242]
	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:175) ~[na:1.8.0_242]
	at java.util.stream.Streams$StreamBuilderImpl.forEachRemaining(Streams.java:419) ~[na:1.8.0_242]
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482) ~[na:1.8.0_242]
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472) ~[na:1.8.0_242]
	at java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150) ~[na:1.8.0_242]
[...]

PR #751, opened 2020-04-26
Fix processing updates with no user info

addo37:
This addresses #749 again. It spawns the EMPTY_USER with empty equivalent arguments and most importantly adds two more tests.


kortov:
I think we can merge this? :)

kortov:
@rubenlagus @addo37 can you merge this?

addo37:
I'd love to, but my guess is that Ruben prefers releasing a new version only if there's a Bot API upgrade or an important bug fix needed. Also, I don't have the right access privileges to merge anything.

PR #765, opened 2020-06-13
Small improvements

Samehadar:
Add AUDIO flag;
Make name 'DEFAULT' public [from now it is available for usage in AbilityExtension;


rubenlagus:
Please, redo against Dev branch

PR #766, opened 2020-06-15
Fix reply flow registration when using ability definitions

addo37:
ReplyFlows don't fully work when defined in an ability. This PR fixes the issue by calling the required Reply::stream function that would return all the attached replies for a particular ReplyFlow.

Also in this PR, we're adding supports for stats in ReplyFlows. Simply, `enableStats` in a ReplyFlow builder and stats would be enabled.


k1ryl:
@addo47 Hello. Could you please provide an example how to define ReplyFlow in Ability?

Issue #768, opened 2020-06-15
Ability command does not triggering in certain cases


Melodeiro:
When i type command with some text on next line, it does not working:
```
/test
text1
text2
```
But if i type something on the first line it is working:
```
/test something
text1
text2
```
Im using 4.9 version, with AbilityBuilder and `name(test)` method.

addo37:
The command mechanism splits the input by ' ' then takes the initial token as the ability name. This won't work with multi-line.

I think this could be enhanced. In the meantime, you can always use a Reply to support this use case. https://github.com/rubenlagus/TelegramBots/wiki/Using-Replies

addo37:
@Melodeiro Can you verify if this is still not addressed? If it is, can you close the issue? Thanks!

Here's the reference to the new methods introduced to customize the way we process text.

https://github.com/rubenlagus/TelegramBots/wiki/Advanced

Melodeiro:
Added custom regex for spaces and newlines, and it is working as i wanted, thank you

PR #780, opened 2020-07-18
Add customization options for command processing

addo37:
This improves the text processing for AbilityBots by supplying three additional parameters:
`getCommandPrefix` - Overriding would allow users to change the prefix
`getCommandRegexSplit` - Overriding would allow users to change how the command is split based on the supplied regex
`allowContinuousText` - Overriding would allow users to process commands as soon as they are matched

I've added the corresponding wiki sections to further expand on these use cases.


addo37:
This PR addresses #768.

PR #782, opened 2020-07-20
Add CommandMessage class

Borumer:
This class adds argument functionality to a given message that a user typed, that also counts as a command


Borumer:
It is a lightweight class that lets developers easily pull out the part of the String that is the command (String), and the part that is the set of parameters (String array), or a parameter (String)

rubenlagus:
Please, do request against DEV

PR #798, opened 2020-08-30
Catch-all clauses should catch Throwable instead of Exception class.

shalupov:
Base class for all exception is Throwable, and it's possible to skip some of them by catching only Exception.
e.g., throwing AssertionError will lead to thread termination

### DIFF
--- a/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/bot/BaseAbilityBot.java
+++ b/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/bot/BaseAbilityBot.java
@@ -599,7 +599,7 @@ public abstract class BaseAbilityBot extends DefaultAbsSender implements Ability
     boolean runSilently(Callable<Boolean> callable, String name) {
         try {
             return callable.call();
-        } catch(Exception ex) {
+        } catch(Throwable ex) {
             log.error(format("Reply [%s] failed to check for conditions. " +
                 "Make sure you're safeguarding against all possible updates.", name));
         }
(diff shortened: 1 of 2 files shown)


addo37 (commented):
I dislike the general idea of catching throwables, but the Reply use case is quite generic that it might warrant this. With these changes, an Error would technically be swallowed.

What use case did you come across?

SandaruKasa:
> Base class for all exception is Throwable, and it's possible to skip some of them by catching only Exception.
> e.g., throwing AssertionError will lead to thread termination

Catching `Throwable` does not sound like a good idea at all honestly. I cannot think of any sane circumstances under which you would even want to catch a [`VirtualMachineError`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/VirtualMachineError.html). If `AssertionError` is the only non-`Exception` `Throwable` that you _actually_ want to catch, maybe `catch(Exception | AssertionError ex)` is what you are looking for?

shalupov:
@addo37 Sorry, missed the first message

> I dislike the general idea of catching throwables

Any known problems with catching throwables?

@SandaruKasa 
I've checked `Error` inheritors and it's quite a few of them. Some of them are very random and could be thrown from anywhere (NotImplementedError? ANTLRError? com.google.common.util.concurrent.ExecutionError? AWTError?)
https://gist.github.com/shalupov/593fcf3240b6739556fadf14b84a1bba

> but the Reply use case is quite generic that it might warrant this. With these changes, an Error would technically be swallowed.
> What use case did you come across?

I was just reading code when debugging why my service stopped. In the end, it was unrelated to TelegramBots.

I remember making this change (Exception -> Throwable) across a bigger project where some exceptions were not correctly logged.

addo37:
I'm fine with the throwable catch here @rubenlagus, but there's code changed outside AbilityBot. I. generally don't advise against catching Throwables except when the use case is generic enough, like the Reply-error-catching.

rubenlagus:
@shalupov Have you seen any case when throwing a `Error` inheritor makes sense to keep running the process? They usually imply bigger problems that the applications won't easily be able to handle on its own and recover

recursiveribbons:
SonarLint does recommend that Throwables not be caught, unless there is a good reason to. Catching Exception is preferred.

PR #802, opened 2020-09-07
Ability webhook action

MouamleH:
The ability now has a `webhookAction` you can use to return an API response instead of using `action`

### DIFF
--- a/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/objects/Ability.java
+++ b/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/objects/Ability.java
@@ -4,11 +4,13 @@ import com.google.common.base.MoreObjects;
 import com.google.common.base.Objects;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.telegram.telegrambots.meta.api.methods.BotApiMethod;
 import org.telegram.telegrambots.meta.api.objects.Update;
 
 import java.util.Arrays;
 import java.util.List;
 import java.util.function.Consumer;
+import java.util.function.Function;
 import java.util.function.Predicate;
 
 import static com.google.common.base.Preconditions.checkArgument;
@@ -34,6 +36,7 @@ import static org.apache.commons.lang3.StringUtils.*;
  *
  * @author Abbas Abou Daya
  */
+@SuppressWarnings("rawtypes")
 public final class Ability {
   private static final Logger log = LoggerFactory.getLogger(Ability.class);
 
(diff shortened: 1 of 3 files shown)


telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/objects/Ability.java
```diff
@@ -63,7 +67,15 @@ private Ability(String name, String info, Locality locality, Privacy privacy, in
...
 
-    this.action = checkNotNull(action, "Method action can't be empty. Please assign a function by using .action() method");
+    this.webhookAction = webhookAction;
+    boolean isWebhook = webhookAction != null;
+
+    if (!isWebhook) {
+      this.action = checkNotNull(action, "Method action can't be empty. Please assign a function by using .action() method");
+    } else {
+      this.action = null;
+    }
```
addo37, line 77: You're enforcing a webhook action on all declared abilities. I haven't looked at your compilation/testing results, but I guess this is why your Travis build failed.

You should rethink how you're going to make the framework accept an action that returns a method. Perhaps this warrants a WebhookAbility that changes the action in a function.

addo37: That also has other implications since Ability is final. Re-architecting that needs to be handled with care.

addo37 (changes requested):
What exactly do we gain by supporting returned BotApiMethods? Are we missing out if we're not supporting them?

Thanks for the effort, but this definitely needs a deeper dive.

telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/objects/Ability.java
MouamleH: "You're enforcing a webhook action on all declared abilities."
I don't get what you mean by this.

addo37: Sorry, I misread. No, we're not enforcing here the webhook action, but I think this may allow users of the long-polling variant to assign a webhook action without assigning an action and have that pass the  pre-conditions. Another caveat is that developers would be able to declare an ability with both actions, yet only one will be acted on. A third caveat would be that the webhook action is not properly checked for null. Unlike the normal action, this wouldn't throw an exception if the webhook action is null. That's because you're using that exact logic to determine whether the Ability is webhook or not. That's bad design. I think we can do better.

What I'm saying here is, relying on whether the user has supplied a webhook action to determine whether the Ability itself targets a Webhook bot seems like a very loose relationship to me. It would be a stronger implication to declare a Webhook Ability (or even push this logic up to the builder!).

I'm here to help, so if anything is unclear, shoot me a DM.

addo37: So, we can do something like:
1 - BaseAbility
2 - Ability
3 - WebhookAbility
4 - AbilityBuilder <- abstract and package-private

Then the Ability variants will have their own builders, extending the AbilityBuilder and performing the separate action setting.

MouamleH: I wanted to go with that suggestion from the beginning but I didn't want to change that much code, fearing it might break so much stuff along the way.
All issues mentioned above will be fixed automatically if we go with that approach. 

Other than that, I really like the idea and will implement it in a separate branch.
And I will be happy to hear your comments/suggestions on it <3.

addo37: Any updates on this PR? A new version is coming soon and if you got something down we could get it merged. Otherwise, it'll have to wait for the next cycle.

MouamleH: Sorry, I've been too busy.
we'd have to wait for the next cycle.

PR #808, opened 2020-10-02
Add InputStream constructor for InputMedia

NordiaKt:


rubenlagus:
Please, add some tests to this change to ensure correctness in the process

rubenlagus:
Actually, caption is an optional field, hence not added to the existing constructors. Going this path would lead to having dozens of constructors really difficult to maintain and use. Hence, only mandatory fields are added.

PR #810, opened 2020-10-02
Be able to access AbilityBot via MessageContext

christianblos:
This pull request makes it possible to access the bot class through the MessageContext object. For instance:

```
public class UuidAbility implements AbilityExtension {

    public Ability createAbility() {
        return Ability.builder()
            .name("uuid")
            .info("Generates a UUID")
            .locality(Locality.USER)
            .privacy(Privacy.PUBLIC)
            .action(ctx -> ctx.bot().silent().send(UUID.randomUUID().toString(), ctx.chatId()))
            .build();
    }

}
```

As you can see in the action method I can now use `ctx.bot().silent()` to send a message.

PR #813, opened 2020-10-05
Be able to add AbilityExtensions in Bot constructor

christianblos:
Currently it's only possible to add ability extensions by defining a method that returns an AbilityExtension. This PR allows to add extensions in the Bot constructor already. This is especially useful when using Spring Dependency Injection.

Together with my other Pull Requests [#810](https://github.com/rubenlagus/TelegramBots/pull/810) and [#812](https://github.com/rubenlagus/TelegramBots/pull/812) you can now create a Bot like this:

```
@Component
public class MyBot extends AbilityBot {

	private final MyBotConfig config;

	public MyBot(MyBotConfig config, List<AbilityExtension> extensions) {
		super(config.getToken(), config.getUsername(), MapDBContext.onlineInstance(config.getDbFile()));
		this.config = config;
		this.extensions.addAll(extensions);
	}

	@Override
	public int creatorId() {
		return config.getCreatorUserId();
	}

}
```

All ability extensions are now registered automatically if they are configured as a Spring Bean.

### DIFF
--- a/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/bot/BaseAbilityBot.java
+++ b/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/bot/BaseAbilityBot.java
@@ -106,7 +106,7 @@ public abstract class BaseAbilityBot extends DefaultAbsSender implements Ability
     private final String botUsername;
 
     // Ability registry
-    protected final List<AbilityExtension> extensions = new ArrayList<>();
+    private final List<AbilityExtension> extensions = new ArrayList<>();
     private Map<String, Ability> abilities;
     private Map<String, Stats> stats;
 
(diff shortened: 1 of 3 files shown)


telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/bot/BaseAbilityBot.java
```diff
@@ -106,6 +106,7 @@
     private final String botUsername;
 
     // Ability registry
+    protected final List<AbilityExtension> extensions = new ArrayList<>();
```
addo37, line 109: I find that exposing this (and the pattern of usage) is not optimal. We could keep this a private list and expose a "addExtensions" method that will add the extension.

```java
public addExtension(List extensions) ....`

addo37 (changes requested):
Again, thanks for the changes.

telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/bot/BaseAbilityBot.java
christianblos: Good point, I also thought about this. Will change it 👍

christianblos:
I updated the wiki and test as well. Note that this PR requires #812 to be merged first! The ExtensionTest is currently failing because the abilities are still registered in the constructor. As soon as #812 is merged and the abilities are loaded later, the test will succeed.

rubenlagus:
Same here, please move to dev branch

christianblos:
Target is now the dev branch

PR #814, opened 2020-10-06
Change Ability name check according to Telegram bot API requirements

Daniil547:
Currently Ability `name`  is checked to contain alpha-numeric characters and no white spaces (beside not being null nor empty).
This doesn't correspond with Telegram API requirements, which are:

> "A command must always start with the '/' symbol and may not be longer than 32 characters. Commands can use latin letters, numbers and underscores." (https://core.telegram.org/bots#commands)

This PR makes underscores allowed to use in Ability `name` and adds check for its length (requirement of leading '/' is irrelevant because Ability `name` isn't a command itself, just a "name" of a command, so to speak; however this '/' has to be taken into account, since it leaves 31 character for the rest of the command)

### DIFF
--- a/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/objects/Ability.java
+++ b/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/objects/Ability.java
@@ -50,9 +50,8 @@ public final class Ability {
 
   @SafeVarargs
   private Ability(String name, String info, Locality locality, Privacy privacy, int argNum, boolean statsEnabled, Consumer<MessageContext> action, Consumer<MessageContext> postAction, List<Reply> replies, Predicate<Update>... flags) {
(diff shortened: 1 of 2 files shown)


addo37 (approved):
Thanks for making these changes, @Daniil547.

zundarik:
Good afternoon. This change is badly needed. An underscore is required in the command name.
We need to make changes to the line checkArgument (isAlphanumeric (name), "Method name can only be alpha-numeric", name);
It does not accept the underscore.

Chase22:
It's funny how "can use underscores" became "must use underscores" in just 3 comments

rubenlagus:
@zundarik Do the pull request against dev branch

Daniil547:
> 
> @rubenlagus:
> @zundarik Do the pull request against dev branch

I was the one who opened the PR, so I suppose that should have been addressed to me. I'll do that.

Issue #770, opened 2020-06-18
Incorrect TelegramLongPollingCommandBot class constructor implementation causing NPE in CommandRegistry.removeUsernameFromCommandIfNeeded(String command) method


vikulin:
Incorrect TelegramLongPollingCommandBot class constructor implementation causing NPE in CommandRegistry.removeUsernameFromCommandIfNeeded(String command) method.
See the code line:

https://github.com/rubenlagus/TelegramBots/blob/1a2e7c9773dbb7d251d30f1f8e3d6538dcd92600/telegrambots-extensions/src/main/java/org/telegram/telegrambots/extensions/bots/commandbot/TelegramLongPollingCommandBot.java#L56

In time the TelegramLongPollingCommandBot contructor creates the instance method this.getBotUsername() returns null because all fields of implementing class have not been initialized yet. Fields initialization runs **after** super.TelegramLongPollingCommandBot  invocation

Suggested fix: add String botUsername parameter in TelegramLongPollingCommandBot constructor passing to CommandRegistry instance.

rubenlagus:
@vikulin Feel free to do a pull request with the fix, as long as it is backward compatible, I'm ok with suggested fix

vikulin:
@rubenlagus , yesm I have created the pull request. See https://github.com/jenkinsci/telegram-notifications-plugin/pull/58

vikulin:
Also it should be corrected accordingly to my suggestion.

rubenlagus:
@vikulin That's a different project... not this one

vikulin:
@rubenlagus , what is the project?

rubenlagus:
@vikulin Probably the one where you created the issue and are writing now? :)

https://github.com/rubenlagus/TelegramBots

vikulin:
@rubenlagus ah, yes. There is still only silence...

rubenlagus:
Should be fixed now

PR #820, opened 2020-10-20
CommandRegistry: Replace botUsername with botUsernameSupplier

### LINKED ISSUE #796: TelegramLongPollingCommandBot can't get correct botUserName
[TelegramLongPollingCommandBot](https://github.com/rubenlagus/TelegramBots/blob/master/telegrambots-extensions/src/main/java/org/telegram/telegrambots/extensions/bots/commandbot/TelegramLongPollingCommandBot.java)

In constructor `public TelegramLongPollingCommandBot(DefaultBotOptions options, boolean allowCommandsWithUsername)`
method `getBotUsername()` returns null if the name is pulled through the class field, from a config-file for example

alexengrig:
Fix #796 and #770.

### DIFF
--- a/telegrambots-extensions/src/main/java/org/telegram/telegrambots/extensions/bots/commandbot/TelegramLongPollingCommandBot.java
+++ b/telegrambots-extensions/src/main/java/org/telegram/telegrambots/extensions/bots/commandbot/TelegramLongPollingCommandBot.java
@@ -53,7 +53,7 @@ public abstract class TelegramLongPollingCommandBot extends TelegramLongPollingB
      */
     public TelegramLongPollingCommandBot(DefaultBotOptions options, boolean allowCommandsWithUsername) {
         super(options);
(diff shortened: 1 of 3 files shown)


rubenlagus:
@alexengrig Can we keep the original constructor but mark it as deprecated? Otherwise it will be a breaking change:

`public CommandRegistry(boolean allowCommandsWithUsername, String botUsername) `

alexengrig:
@rubenlagus Return the original constructor. Need to mark it as deprecated?

rubenlagus:
@alexengrig if the old constructor fix issue #770 no need, otherwise, let mark is as deprecated

rubenlagus:
@alexengrig In that case, better to deprecate it and add something to the docs to state that situation. I'll make sure it is removed in next major version

alexengrig:
@rubenlagus Done.

PR #821, opened 2020-10-20
Add codecov plugin

alexengrig:
[Codecov plugin](https://github.com/alexengrig/codecov-maven-plugin).


alexengrig:
Need to start Travis CI build.

rubenlagus:
What's the benefit of this plugin against the default travis config that we had?

alexengrig:
@rubenlagus To be honest, there is only run with maven 😄.

PR #822, opened 2020-10-25
Enum for ChatMember statuses and Message#hasCaption

nordiauwu:


telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/objects/ChatMember.java
```diff
@@ -189,4 +189,13 @@ public String toString() {
                 ", customTitle=" + customTitle +
                 '}';
     }
+
+    public enum Status {
```
rubenlagus, line 193: Please move this to a separate file

nordiakt: +

rubenlagus (approved):

rubenlagus:
Happy to merge this change but please extract the status class to a different class and name it with something that indicate it refers to chat member statuses.

nordiauwu:
Outdated

PR #832, opened 2020-11-12
Fix webhook url and nullpointer

samat-rt:
When we send webhook url to telegram, we will get only server path. 
Also add default value to DefaultWebhook.internalUrl


rubenlagus:
Changes applied manually

PR #835, opened 2020-11-24
Inject bot instance into the reply update consumer

Chase22:
Also added deprecated methods to still allow the old behaviour, discarding the bot reference in that case.

The condition could have a bot instance injected as well, which would require rewriting the Flag register though.


addo37:
@rubenlagus This is good to go; thanks Chase.

PR #843, opened 2020-12-09
wiki: fix code in ability extension example

Dhina17:


addo37:
Ouch, I hope this was just a faulty PR from before.

This is good to go.  @rubenlagus

Issue #850, opened 2021-01-02
check token and username on blank string


centralhardware:

centralhardware:
PR already merged

PR #851, opened 2021-01-02
check that username or token are not null or empty

centralhardware:
to prevent can't remove/delete webhook when token or username are not present

https://github.com/rubenlagus/TelegramBots/issues/850

### DIFF
--- a/telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/TelegramBotsApi.java
+++ b/telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/TelegramBotsApi.java
@@ -2,10 +2,7 @@ package org.telegram.telegrambots.meta;
 
 import org.telegram.telegrambots.meta.api.methods.updates.SetWebhook;
 import org.telegram.telegrambots.meta.exceptions.TelegramApiException;
-import org.telegram.telegrambots.meta.generics.BotSession;
-import org.telegram.telegrambots.meta.generics.LongPollingBot;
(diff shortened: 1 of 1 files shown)


telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/TelegramBotsApi.java
```diff
@@ -88,4 +91,16 @@ public void registerBot(WebhookBot bot, SetWebhook setWebhook) throws TelegramAp
...
         }
     }
+
+    /**
+     * checking that the username and token are presented
+     * @param bot bot to register
+     * @return true if username or token is empty or null
+     */
+    public boolean checkBotCredentials(TelegramBot bot){
+        return bot.getBotUsername().isEmpty()   ||
```
Chase22, line 101: This is going to cause a NullPointerException if the username is null

telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/TelegramBotsApi.java
```diff
@@ -88,4 +91,16 @@ public void registerBot(WebhookBot bot, SetWebhook setWebhook) throws TelegramAp
...
+                bot.getBotToken().isEmpty()     ||
```
Chase22, line 102: Same here

Chase22 (changes requested):

Chase22 (approved):

rubenlagus:
Merged separatelly via manual commit

PR #862, opened 2021-01-30
Fix Gradle 7.0 compatibility

ekiauhce:
The usage of `compille` configuration has been discouraged since Gradle 3.4 ([link](https://docs.gradle.org/current/userguide/upgrading_version_5.html#changes_6.0)).
And this configuration will fail with an error in Gradle 7.0. So we need to use `implementation` instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,12 +31,14 @@ Just import add the library to your project with one of these options:
     </dependency>
 ```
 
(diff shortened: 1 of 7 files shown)


costalfy (commented):
ok for me

PR #874, opened 2021-03-12
Fix typos

aNNiMON:
- Fixed PATH in `EditChatInviteLink` and `RevokeChatInviteLink`
 - Fixed `getUpdatesTimeout` not sets in `DefaultBotOptions`. This causes to use short-polling instead of long-polling.  
   See https://core.telegram.org/bots/api#getupdates:  
   > Timeout in seconds for long polling. Defaults to 0, i.e. usual short polling. Should be positive, short polling should be used for testing purposes only.


rubenlagus:
Thanks for this, released

Issue #892, opened 2021-04-07
This could be handled in a bit more efficient way (at least telling the root cause for such behaviour)


slavino:
https://github.com/rubenlagus/TelegramBots/blob/4882040822d202e146471d6196e3a40babbdb51a/telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/methods/updates/GetUpdates.java#L91

![image](https://user-images.githubusercontent.com/1584034/113854825-9af0d200-979f-11eb-9344-8c4028e49dc8.png)

![image](https://user-images.githubusercontent.com/1584034/113855173-076bd100-97a0-11eb-8b25-fe2de5a20794.png)

Maybe just handling it a bit differently on line 270 would make huge difference. I was investigating whether it is timeout issue and only accidentally checked the message itself - mea culpa, I know... but could have been much smoother if these few characters contained in log.error (concatenated localized message + apiResponse). What would you say to such small change?

PR #893, opened 2021-04-07
Update DefaultBotSession.java

slavino:
https://github.com/rubenlagus/TelegramBots/issues/892


rubenlagus:
@slavino What about overriding the actual getMessage method form the TelegramApiRequestException (or any other exception in the API to something like (yes, looks like initially I override `ToString` incorrectly:

```
        @Override
        public String getMessage() {
            if (super.getApiResponse() == null) {
                return super.getMessage();
            } else if (super.getErrorCode() == null) {
                return super.getMessage() + ": " + super.getApiResponse();
            } else {
                return super.getMessage() + ": [" + super.getErrorCode() + "] " + super.getApiResponse();
            }
        }
```

slavino:
Sure, go ahead ;)

PR #895, opened 2021-04-08
telegrambots-spring-boot-starter: add support for TelegramLongPollingCommandBot

centralhardware:
automatic register BotCommand in TelegramLongPollingCommandBot

it's allow  to use ideomatic  spring DI in BotCommand classes


telegrambots-spring-boot-starter/src/test/java/org/telegram/telegrambots/starter/TestTelegramBotStarterRegistrationHooks.java
```diff
@@ -11,6 +11,7 @@
 import org.telegram.telegrambots.meta.exceptions.TelegramApiException;
 import org.telegram.telegrambots.meta.generics.BotSession;
 import org.telegram.telegrambots.meta.generics.LongPollingBot;
+import org.telegram.telegrambots.starter.Annotation.AfterBotRegistration;
```
UnAfraid, line 14: Is that necessary?

UnAfraid:
I've nothing against the idea, but the implementation kinda forces you to use the long pooling mechanism, while this is implemented in spring extension which is by nature is a web serving framework.
I just think it makes no sense the way command bot implementation works to be directly integrated that way.

Issue #494, opened 2018-07-18
How can i send photo via sendPhoto method in the async way?


CheremisinVladislav:
Hey, I have built a bot which replies with images when a command is passed to it.
When  messaged are being sent from two devices  simultaneously, images are sent in turns.First to one device. Next to another. Does it mean that longpolling handle a request in the sync way or i don't setup it proper way?

FRedEnergy:
Any updates on this one? It would be usefull to have async version for uploading documents, videos, photos, etc, built in and not having to setup your own async queue.

rubenlagus:
New sendXXXXXAsync method could be added without issues, a pull request would be appreciate through.

genjiezio:
I 'll try to work on this，any advice is appeciated.

genjiezio:
> New sendXXXXXAsync method could be added without issues, a pull request would be appreciate through.

Excuse me,  I find the ***excuteAsync()*** method in ***DefaultAbsSender.java***, does it means this issue should already have been closed ?

genjiezio:
> New sendXXXXXAsync method could be added without issues, a pull request would be appreciate through.
Excuse me, I find the **excuteAsync**  wrote by ***CompletableFuture*** in the **DefaultAbsSender.java** when I try to merge The Async method i wrote by ***FutureTask***,  does this means that this issue should have been closed in 2020?

PR #904, opened 2021-04-23
Modify DefaultAbsSender.java for issue #494

genjiezio:
#494 
I try to use the ThreadPool (**newCachedThreadPool**) to excute the sendphoto Async.
And it has a good performance , which can be seen in this url:
https://github.com/genjiezio/ImageRepository/blob/main/BSY8UISCOQD752OGHMO91MF.png

### DIFF
--- a/telegrambots/src/main/java/org/telegram/telegrambots/bots/DefaultAbsSender.java
+++ b/telegrambots/src/main/java/org/telegram/telegrambots/bots/DefaultAbsSender.java
@@ -150,7 +150,7 @@ public abstract class DefaultAbsSender extends AbsSender {
     public void Async_execute(SendPhoto sendPhoto){
         executor.submit(()->{
             try{
-                new Thread (String.valueOf(execute(sendPhoto)));
+                execute(sendPhoto);
             }catch (Exception e){
                 e.printStackTrace();
             }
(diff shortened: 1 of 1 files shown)


genjiezio:
So, after I pull a request for this method, the Coverage told that  a **excuteAsync** had been created in this .java file, which use **CompletableFuture**,  which i didn't notice.
So why not close the issue #494 ?
Does this means it still  need to be improve?

rubenlagus:
Please do the request against dev branch.

And would help having a simple Unit Test for the new method 😅

genjiezio:
> Please do the request against dev branch.
> 
> And would help having a simple Unit Test for the new method 😅

Yeah, i'd like to do that. But actually, now I find you had wrote a **excuteAsync** in the 2020 which i didn't notice when i wrote this method, does this means the issue was solved by you in 2020?  Then why still keep this issue?

For the Unit Test, actually, i think it difficult to test use Junit. I test the method by manually send signals concurrently to the bot but it's still can't judge whether it really excute asynchronously because the excute time is only about 40ms. 

No matter how, i'll try to write a Unit test as much as possiable. I do this request because I can't get yours advice from the issue page since i commented in the issue page **#494** in March, so i want to get your attention in this pr, I really want to help in this issue.

>

rubenlagus:
Correct, the problem was fixed with the addition of the async methods, so no need to add any other methods here

`public CompletableFuture<Message> executeAsync(SendPhoto sendPhoto)`

Issue #514, opened 2018-08-21
how to use inline model?


boy-good:
how to set inline model,like this button..

![image](https://user-images.githubusercontent.com/39320533/44385724-d557c300-a552-11e8-847c-b37bc5a67ca0.png)

Clevero:
Those aren't Inline Buttons.

I think you mean custom keyboards. How to use them can be read in the FAQ
https://github.com/rubenlagus/TelegramBots/wiki/FAQ#how_to_use_custom_keyboards

mymike00:
but how can I use those with an AbilityBot ?

PR #905, opened 2021-04-23
Inline model example for issue  #514

nicetryzz:
A inline model example is added, you can run Main and  use /inline command check with robot "InlineModelBot".


rubenlagus:
Why are you adding samples in the library repository? 

Please add them to the wiki

Issue #803, opened 2020-09-09
Cannot get slow mode delay


Stev4iKFeeD:
I cannot get correct slow mode delay. It's always null. I tried to promote my bot to admins and turned off privacy mode in botfather but it didn't help
![image](https://user-images.githubusercontent.com/60626324/92662768-33a59900-f308-11ea-9777-c92423d1d1d2.png)
![image](https://user-images.githubusercontent.com/60626324/92662939-9f880180-f308-11ea-9f9b-690163b5e4b3.png)
What should I do?

Thanks

PR #908, opened 2021-04-25
answer to issue 803

thousandsofbug:
#803 
> I cannot get correct slow mode delay. It's always null. I tried to promote my bot to admins and turned off privacy mode in botfather but it didn't help
> image
> image
> What should I do?
> 
> Thanks
USE getChat(//your chat id there) then you can get slow mode delay
image
image


rubenlagus:
How is this solving the issue? It is only adding new test cases

thousandsofbug:
I think it is a misoperation, reply to him than PR

Issue #900, opened 2021-04-19
Just a maven 3.5.+ notice on modules version inheritance


slavino:
Improvement proposal:

just a hint on releasing (if needed)
https://stackoverflow.com/questions/10582054/maven-project-version-inheritance-do-i-have-to-specify-the-parent-version

xcjusuih:
Do you means that we can define some {version} in parent pom.xml and using it directly in child module so when a new version is released, only need to change the version defined in parent pom.xml?

PR #910, opened 2021-04-25
Update pom.xml

xcjusuih:
Change pom.xml in telegrambot as a case for #900.


rubenlagus:
The change is just removing the parts necessary to deploy the library to Maven and the enforcer.

PR #911, opened 2021-04-26
Update README.md

slavino:
added exclusion to avoid dependencies if newer spring boot is used in the user project


rubenlagus:
This is not part of the basic usage step for the library to be added to the readme.

Issue #299, opened 2017-09-11
Need to have possibility to check status of connection in long polling mode


dburnazyan:
When I use Long Polling Bot I need to have possibility check if all work fine: is bot is connected to server or something goes wrong.

Clevero:
class BotSession has a method called ```isRunning(): boolean````
https://github.com/rubenlagus/TelegramBots/blob/383b48cddd70af742782493d2e32d53f8fadcd4b/telegrambots-meta/src/main/java/org/telegram/telegrambots/generics/BotSession.java#L27

dburnazyan:
But this method returns true even if connection is lost

dburnazyan:
method run() of ReaderThread class looping until 'running' is true

https://github.com/rubenlagus/TelegramBots/blob/383b48cddd70af742782493d2e32d53f8fadcd4b/telegrambots/src/main/java/org/telegram/telegrambots/updatesreceivers/DefaultBotSession.java#L178

running becomes false only when stop method of DefaultBotSession class is calling

So when telegram server is not accessible ReaderThread continues looping and just generates logs. But isRunning() still returns true because ReaderThread has't changed parameter 'running' to false

dburnazyan:
I suppose that there should be another method like 'isConnectionFine' or something like this and when httpclient request throws some exception while execution this method should return false.

Chase22:
The DefaultBotSession has a try-catch where the actual request is executed

https://github.com/rubenlagus/TelegramBots/blob/383b48cddd70af742782493d2e32d53f8fadcd4b/telegrambots/src/main/java/org/telegram/telegrambots/updatesreceivers/DefaultBotSession.java#L196

We could add either just a boolean if that try-catch was sucessfull, add a callback if it didn't or make an enum with a few more statuses depending on why it failed.

genjiezio:
I'd like to add a new "isConnectMethod" to return a boolean value about connection status, and won't set isrunning to false directly.

PR #922, opened 2021-05-18
Add new Feature "isConnectFine" in DefaultBotSession.java linked to Issue #299

genjiezio:
A new mothod **isConnectFine** was added. Which just gives a judgement about connection status according to the number of Continuous Exception, and doesn't set **running** to false directly. #299


rubenlagus:
Maybe the number of errors could be done configurable via BotOptions? Instead of enforcing 5?

PR #937, opened 2021-07-03
Set the proper fields for forceReply method in silent sender

addo47:


rubenlagus:
`setForceReply(true)` will be defaulted in next version to true, merging just for the selective flag.

PR #959, opened 2021-08-19
Simple-Example.md - Added missing /stats command

jonasx10j:
/stats was missing from the default commands list.


addo47:
Good to go, long overdue @rubenlagus 

Thanks @jonasx10j

PR #1010, opened 2021-12-09
PR for issue 935

kunzhang0516:
added cancel method in bot DefaultAbilities and unit test

### DIFF
--- a/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/bot/DefaultAbilities.java
+++ b/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/bot/DefaultAbilities.java
@@ -75,6 +75,7 @@ public final class DefaultAbilities implements AbilityExtension {
   public static final String DEMOTE = "demote";
   public static final String UNBAN = "unban";
   public static final String BACKUP = "backup";
+  public static final String CANCEL = "cancel";
   public static final String RECOVER = "recover";
   public static final String COMMANDS = "commands";
   public static final String REPORT = "report";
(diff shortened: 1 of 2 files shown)


telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/bot/DefaultAbilities.java
```diff
@@ -236,6 +237,25 @@ public Ability backupDB() {
...
+     * The ability to cancel operation.
+     * User can use this method to cancel the current session and delete related chat.
+     * Usage: <code>/cancel</code>
+     *
+     * @return the ability to cancel operation
+     */
+    public Ability cancel() {
+        return builder()
+                .name(CANCEL)
+                .info("Cancel current operation.")
```
addo47, line 250: Don't supply info as that would make it be reported in the commands report ability. Do add it to the wiki where appropriate.

Without info set, this ability will still be reported using the /commands operation. But, /report won't.

telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/bot/DefaultAbilities.java
```diff
@@ -236,6 +237,25 @@ public Ability backupDB() {
...
+                .privacy(PUBLIC)
+                .locality(ALL)
+                .action(ctx -> {
+                    bot.db.<Long, Integer>getMap("user_state_replies").remove(ctx.chatId());
```
addo47, line 254: Please use STATES variable in ReplyFlowBuilder. Same for the test code.

addo47 (changes requested):

PR #1011, opened 2021-12-09
Issue 834

nubdub:
Add constant DEFAULT to AbilityExtension.java and getter method

### DIFF
--- a/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/util/AbilityExtension.java
+++ b/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/util/AbilityExtension.java
@@ -5,8 +5,14 @@ package org.telegram.abilitybots.api.util;
  */
 public interface AbilityExtension {
 
+    // https://github.com/rubenlagus/TelegramBots/issues/834
     String DEFAULT = "default";
 
+    /**
+     * https://github.com/rubenlagus/TelegramBots/issues/834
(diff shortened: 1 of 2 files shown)


Chase22:
This should be a public static final without a getter

telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/util/AbilityExtension.java
```diff
@@ -4,4 +4,17 @@
...
+
+    /**
+     * https://github.com/rubenlagus/TelegramBots/issues/834
+     * Method for accessing AbilityExtension.DEFAULT
+     * @return String "default"
+     */
+    default String getDefault() {
+        return AbilityExtension.DEFAULT;
+    }
+
```
addo47, line 19: I don't like the idea that there's two declarations to this variable (here + in base). We can switch up the DEFAULT variable to become public. I don't see a problem in widening that.

addo47 (changes requested):

telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/util/AbilityExtension.java
addo47: On a second thought, we can keep this here and remove the one in base ability bot. That class already extends AbilityExtension and people who're implementing the default outside the bot's class will benefit from this localization.

Also, ye this is definitely a static variable with a static method.

PR #1046, opened 2022-03-25
Pull up method to reduce duplication

raymanlk:
DeserializeResponse method was duplicated twice in Delete and Set, so was pulled upto an intermediate class BooleanBotApiMethod as there was a third class using deserializeResponse which had a different signature, hence the intermediate class.


rubenlagus:
That same method is use in dozens of other places as well, why only refactor one of them?

PR #1047, opened 2022-03-25
Extracted class

raymanlk:
Mix of aspects was seen in CopyMessage.class . Methods which did not meet responsibility was moved into a new class.


rubenlagus:
Don't see what's the benefit of this change

PR #1055, opened 2022-04-18
Change the format of reply exception log.

LamGC:
Output exception details in exception handling to help developers troubleshoot problems. At the same time, by checking whether the debug level is enabled, too many output logs in the production environment can be avoided. (I guess the original design was for this reason)

After the change, when the debug level is not enabled, the log format is no different from that before. When the debug level is enabled, exception stack information is output to help developers check for problems.

#### Previous format / Debug level not enabled
```
[12:10:56.253 ERROR][o.t.a.api.bot.BaseAbilityBot][Bot Telegram Executor]: Reply [null] failed to check for conditions. Make sure you're safeguarding against all possible updates.
```
#### Debug level enabled
```
15:08:47.369 [main] ERROR org.telegram.abilitybots.api.bot.BaseAbilityBot - Reply [null] failed to check for conditions. Make sure you're safeguarding against all possible updates.
java.lang.RuntimeException: Throwing an exception inside the reply conditions (flags)
	at ......
```


addo47:
This is ok. @rubenlagus

PR #1067, opened 2022-05-14
Add a function for converting Consumer to BiConsumer / fix #1066

### LINKED ISSUE #1066: AbilityBot reply example in Wiki does not compile with telegrambots-abilities v6.0.1
Source: https://github.com/rubenlagus/TelegramBots/wiki/Using-Replies#standalone-reply

Code:
```java
public Reply sayYuckOnImage() {
  // getChatId is a public utility function in rg.telegram.abilitybots.api.util.AbilityUtils
  Consumer<Update> action = upd -> silent.send("Yuck", getChatId(upd)); 
  
  return Reply.of(action, Flag.PHOTO)
}
```

image

I know it's hard to keep all the docs up to date!

montpelllier:
Make the use of `Reply.of()` in previous code and Wiki example available. Try fixing #1066.

### DIFF
--- a/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/bot/BaseAbilityBot.java
+++ b/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/bot/BaseAbilityBot.java
@@ -660,7 +660,7 @@ public abstract class BaseAbilityBot extends DefaultAbsSender implements Ability
 
     boolean filterReply(Update update) {
         return replies.stream()
-                .filter(reply -> runSilently(() -> reply.isOkFor(update), reply.name()))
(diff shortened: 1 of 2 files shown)


montpelllier:
oh, I make a mistake. The edf21e5fe8ab6a937306b3efce76398a06cbd6eb is not what I want to pull in this request

Chase22:
Wouldn't it be a better idea to update the wiki instead? The way to go is to use a BiConsumer and the old way of using Consumers is no longer supported since the 6.0 release. Readding that only add additional complexity imo

rubenlagus:
Agree with @Chase22 , let´s update it instetad

PR #1071, opened 2022-05-21
Remove function isOkFor to fix #960

### LINKED ISSUE #960: Abilitybot - File as reply to message doesn't trigger default ability.
Code:
```
package com.company;

import org.telegram.abilitybots.api.bot.AbilityBot;
import org.telegram.abilitybots.api.objects.Ability;

import static org.telegram.abilitybots.api.objects.Locality.ALL;
import static org.telegram.abilitybots.api.objects.Privacy.PUBLIC;

public class BotA extends AbilityBot {

    public BotA() {
        super(Constants.BOT_TOKEN, Constants.BOT_USERNAME);
    }

    @Override
    public long creatorId() {
        return Constants.CREATOR_ID;
    }

    public Ability Default() {
        return Ability.builder()
                .name(DEFAULT)
                .privacy(PUBLIC)
                .locality(ALL)
                .input(0)
                .action(ctx -> {
                    System.out.println("PRINT : " +ctx.update().toString());
                })
                .build();
    }

    public Ability StartCMD() {
        return Ability
                .builder()
                .name("start")
                .info("Start info.")
                .locality(ALL)
                .privacy(PUBLIC)
                .action(ctx -> {
                    silent.send("Welcome to " + Constants.BOT_USERNAME + ", type /help for additional info.\n\nBot owned by: " + Constants.BOT_OWNER_TAG, ctx.chatId());
                })
                .build();
    }

}

```
**What's supposed to happen:**

- Print out "PRINT : " +ctx.update().toString() 
- There is only 1 additional ability, so everything else other than /start (+ other default commands) should end up in the default one

**What works:**
-  sending text
-  sending the file alone
-  sending text reply to text message
-  All there will print "PRINT : " +ctx.update().toString() 
Expected output (send normal text)

**What doesn't work:**

- Sending file as reply to text won't print anything.
File as reply to text
File sent as reply to text message

- Sending file as reply to file will print error into logger: `ERROR BaseAbilityBot:676 - Reply [null] failed to check for conditions. Make sure you're safeguarding against all possible updates.` 
file as reply to file
File as reply to file_b

montpelllier:
Remove function isOkFor so the bot can take action normally on replies with document. The bug has been fixed in Java 9 according to [here](https://bugs.openjdk.java.net/browse/JDK-8044546), so the function may be useless.

### DIFF
--- a/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/bot/BaseAbilityBot.java
+++ b/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/bot/BaseAbilityBot.java
@@ -660,7 +660,7 @@ public abstract class BaseAbilityBot extends DefaultAbsSender implements Ability
 
     boolean filterReply(Update update) {
         return replies.stream()
-                .filter(reply -> runSilently(() -> reply.isOkFor(update), reply.name()))
+                .filter(reply -> false)
                 .map(reply -> runSilently(() -> {
                     reply.actOn(this, update);
                     updateReplyStats(reply);
--- a/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/objects/Reply.java
+++ b/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/objects/Reply.java
@@ -52,12 +52,6 @@ public class Reply {
     return Reply.of(action, newArrayList(conditions));
   }
 
-  public boolean isOkFor(Update update) {
-    // The following variable is required to avoid bug #JDK-8044546
-    BiFunction<Boolean, Predicate<Update>, Boolean> stateAnd = (state, cond) -> state && cond.test(update);
-    return conditions.stream().reduce(true, stateAnd, Boolean::logicalAnd);
-  }
-
   public void actOn(BaseAbilityBot bot, Update update) {
     action.accept(bot, update);
   }


montpelllier:
fix #960

telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/bot/BaseAbilityBot.java
```diff
@@ -660,7 +660,7 @@ private void updateUserId(User oldUser, User newUser) {
 
     boolean filterReply(Update update) {
         return replies.stream()
-                .filter(reply -> runSilently(() -> reply.isOkFor(update), reply.name()))
```
addo47, line 663: The title of this PR is misleading. You are indeed removing the function which is no longer required, but you're changing the behavior of this pipeline.

The reply must be checked, its flags (Predicates of type Predicate<Boolean>) pass.

telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/objects/Reply.java
```diff
@@ -52,12 +52,6 @@ public static Reply of(BiConsumer<BaseAbilityBot, Update> action, Predicate<Upda
     return Reply.of(action, newArrayList(conditions));
   }
 
-  public boolean isOkFor(Update update) {
-    // The following variable is required to avoid bug #JDK-8044546
-    BiFunction<Boolean, Predicate<Update>, Boolean> stateAnd = (state, cond) -> state && cond.test(update);
-    return conditions.stream().reduce(true, stateAnd, Boolean::logicalAnd);
```
addo47, line 58: Line 57 to be removed, then inline on 58.

addo47 (changes requested):

PR #1075, opened 2022-05-29
Add a validator for markdown message

### LINKED ISSUE #705: HTML and Markdown validator
Hi,

I faced with the problem when i use broadcasting to the users via telegram bot and could't find any specific markdown and html validator based on telegram api.

Will be it be useful to create validation by this requirements:
https://core.telegram.org/bots/api#html-style

or could you please provide existed solution?

Thanks

SUSTechDXiang:
fix #705
I write a validator for markdown style message. It will filter out invalid markdown style messages and send normal messages instead. So that it will not receive an exception.


rubenlagus:
@SUSTechDXiang can some unit tests be added for the validator?

PR #1082, opened 2022-06-06
Refactoring method implementation public abstract T deserializeRespon…

timursergeevichname:
Refactoring method implementation 
```java 
public abstract T deserializeResponse(String answer) throws TelegramApiRequestException;
```

Moved the common code to 
```java 
public T deserializeResponseDefault(String answer, String exceptionMessage) throws TelegramApiRequestException {
        try {
            ApiResponse<T> result = OBJECT_MAPPER.readValue(answer,
                    new TypeReference<ApiResponse<T>>(){});
            if (result.getOk()) {
                return result.getResult();
            } else {
                throw new TelegramApiRequestException(exceptionMessage, result);
            }
        } catch (IOException e) {
            throw new TelegramApiRequestException("Unable to deserialize response", e);
};
```


timursergeevichname:
Opps!!!

PR #1086, opened 2022-06-16
Wiki: fixed checkGlobalFlags method in readme

LegendaryZer0:
Hi! In reading the documentation, I noticed that Flag.PHOTO returns the Flag type, not bool, so i fixed wiki


addo47:
Good to go @rubenlagus

PR #1124, opened 2022-09-25
Remove dependency on org.json:json

recursiveribbons:
It is no longer needed and is just an extra vulnerability vector


rubenlagus:
Rolled back due to build isues

recursiveribbons:
An issue on Jenkins? I'm curious what Jenkins does because mvn package works

rubenlagus:
Ignore it

PR #1135, opened 2022-10-23
Create SECURITY.md

tg-bot543:


recursiveribbons:
This one also doesn't make sense like #1136

PR #1143, opened 2022-11-04
Reworked some of the old comments in javadoc style

AminevRamil:
The work is not finished yet, but I would like to hear maybe a positive assessment, so that I understand that it is worth doing this further.

Thus it is possible to look at comments using IDE hints.
image

I also did a little rearrangement of the class comments to make them look better in IDE tooltips.
image


EpipremnumPinnatum:
I would like to have the javadoc in place. It would help me a lot. I'll maybe also try to write some on it. However, should private fields also be documented or only public classes/methods?

Issue #1144, opened 2022-11-05
Support for Spring Boot 3


recursiveribbons:
Spring Boot 3 will release in November 2022. It has [multiple breaking changes](https://spring.io/blog/2022/05/24/preparing-for-spring-boot-3-0) including a Java 17 and Jakarta EE 9 requirement. I have been experimenting with this over on [a fork](https://github.com/recursiveribbons/TelegramBots/tree/java-11-spring-boot-3), with moving everything to Java 11 and the Spring Boot Starter to Java 17, and Jakarta EE 9.

Polve:
With spering boot 3.2 (just released) there is another problem: the wrong jackson library gets pulled in: 
jackson-module-jaxb-annotations  in place of jackson-module-jakarta-xmlbind-annotations

recursiveribbons:
@Polve probably time to just fork this repo, there's no sign the maintainer is gonna listen to issues

Haarolean:
I'm running this with boot 3.2.1 perfectly without any issues tho

PR #1155, opened 2022-11-28
Spring Boot 3 support

recursiveribbons:
This PR addresses issue #1144 

- Support Spring Boot 3 and Jakarta EE 9
- Switch the release version of Java to 11 for better support of JDK 17
- Switch the release version of telegrambots-spring-boot-starter to Java 17

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,7 @@
         <jackson.version>2.14.0</jackson.version>
         <json.version>20220924</json.version>
         <slf4j.version>2.0.3</slf4j.version>
-        <jakarta.annotation.version>9.1.0</jakarta.annotation.version>
+        <jakarta.version>9.1.0</jakarta.version>
         <lombok.version>1.18.24</lombok.version>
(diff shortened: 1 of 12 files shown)


lex-em:
I think TelegramBots should support both versions of spring boot (2.* and 3.*), should there be 2 branches with different versions (6.3.* and 6.4.*)?

rubenlagus:
As discussed in the chat, planning to add this spring3 upgrade to the next major version together with a big refactor.

@lex-em There will be support for 2.X temporary but support will be reduced to:

1. Securities updates
2. Telegram API updates

patbaumgartner:
A current workaround which works is to use @Import(org.telegram.telegrambots.starter.TelegramBotStarterConfiguration.class) in your app.

I still would suggest having `[META-INF/spring.factories]` for the mean time with a [META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports] side by side. This should already do the job to support both, Spring Boot 2 & 3 at the same time.

PR #1167, opened 2022-12-31
Pass token in constructor

Chase22:
Currently in the backwards compatible version. The constructors without botToken should be removed in the future


telegrambots/pom.xml
```diff
@@ -63,7 +63,7 @@
 
     <properties>
         <java.version>11</java.version>
-        <maven.compiler.release>8</maven.compiler.release>
+        <maven.compiler.release>11</maven.compiler.release>
```
rubenlagus, line 66: As discussed, let's maintain this as 8 for now until next mayor version

mrFloony:
Hi, bring back the constructors without botToken.
I am using the library in JavaEE (EJB 3) in a beans, and beans need a default constructor (without botToken). 
You also need to return the getBotToken method from @deprecated for it to work correctly in beans.

PR #1184, opened 2023-01-16
Fixes

timursergeevichname:


telegrambots/src/main/java/org/telegram/telegrambots/bots/DefaultAbsSender.java
```diff
@@ -584,13 +584,11 @@ public Boolean execute(SetChatPhoto setChatPhoto) throws TelegramApiException {
...
             builder.addTextBody(SetChatPhoto.CHATID_FIELD, setChatPhoto.getChatId(), TEXT_PLAIN_CONTENT_TYPE);
-            if (setChatPhoto.getPhoto() != null) {
-                InputFile photo = setChatPhoto.getPhoto();
-                if (photo.getNewMediaFile() != null) {
-                    builder.addBinaryBody(SetChatPhoto.PHOTO_FIELD, photo.getNewMediaFile());
-                } else if (photo.getNewMediaStream() != null) {
-                    builder.addBinaryBody(SetChatPhoto.PHOTO_FIELD, photo.getNewMediaStream(), ContentType.APPLICATION_OCTET_STREAM, photo.getMediaName());
-                }
+            InputFile photo = setChatPhoto.getPhoto();
+            if (photo.getNewMediaFile() != null) {
```
Chase22, line 588: I'm very sure this can cause a NPE in java. If photo is null in this situation

Chase22 (changes requested):

PR #1187, opened 2023-01-21
fix - change mimeTypeVariable name in Animation object

kimoror:
In other objects (Document, Audio, and others), the objects are called mimeType. It would be cool if everyone had the same name. For example, it makes it easier to use the mapstruct library


rubenlagus:
Applied this by keeping backward compatibility

PR #1196, opened 2023-02-27
Spring configuration

recursiveribbons:
Adding new spring boot config while keeping the old spring.factories. This should provide basic support for both Spring Boot 2 and 3. May still break if you use the wrong Jakarta versions.

> I still would suggest having `[META-INF/spring.factories]` for the mean time with a [META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports] side by side. This should already do the job to support both, Spring Boot 2 & 3 at the same time.
> 
> _Originally posted by @patbaumgartner in https://github.com/rubenlagus/TelegramBots/issues/1155#issuecomment-1445306344_


yvasyliev:
When will it be merged? Looking forward to Spring Boot 3 support 😊

patbaumgartner:
@rubenlagus could you please merge it? It supports Spring Boot 2 & 3. Thank you!

patbaumgartner:
@rubenlagus Is there a way to help you with making your starter ready for Spring Boot 3.1? I would be happy to support you with the work.

rubenlagus:
This should be completed now

PR #1220, opened 2023-05-12
Update Getting-Started.md

iamgul:
Update Getting-Started.md to show the use of Constructor to pass bot-token instead of using the deprecated getBotToken() method.


rubenlagus:
Docs moved to https://rubenlagus.github.io/TelegramBotsDocumentation/telegram-bots.html

PR #1224, opened 2023-05-21
set false as default value for Use#ispremium

### LINKED ISSUE #1223: User#getIsPremium return null instead of false
**Describe the bug**

User#getIsPremium return null instead of false

**Expected behavior**

User#getIsPremium return false when user doesn't have premium subscription

centralhardware:
fix #1223


rubenlagus:
Is premium is an optional field in Telegram Bots API, hence it can be Null. It's a valid value for optional fields: https://core.telegram.org/bots/api#user

centralhardware:
I know about telegram bot api. But I still don't understand why library must have null for false, even if whatever else do it.

centralhardware:
Null is a huge problem for Java and it's be better to don't use null when we can don't use it.

PR #1261, opened 2023-08-14
Command bot update: user activity feature

loolzaaa:
Add user activity feature for both types (long polling and webhook) of command bot:

- Create parametrized `CommandState<T>` that handles state of the command for user between messages
- Create `UserActivity` class that handles user-parametrized `CommandState<T>` and user last activity date and time
- Create `UserActivityHandler` interface with default implementation. Used as user activity repository (Map in default implementation)
- Add new method for `IBotCommand` interface for process stateful commands
- Deprecate `BotCommand#execute(AbsSender absSender, User user, Chat chat, String[] arguments)` method because of lost access for full `Message` object
- Add methods to `CommandRegistry` for start/continue stateful commands
- Refactor `TelegramLongPollingCommandBot` and `TelegramWebhookCommandBot`
- Add more tests and update javadocs
- Add wiki page


loolzaaa:
It looks like the maintainer doesn't have the time to support this library.

I had to create my own that would solve the problem described here: https://github.com/loolzaaa/tgbot4j

Wiki: https://github.com/loolzaaa/tgbot4j/wiki/Command-State-Bot

PR #1265, opened 2023-08-23
Add support for Spring Boot 3+

homich1991:
In accordance to [documentation on migration](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide#auto-configuration-files) to SpringBoot 3 autoconfiguration file, from now should be placed slightly different.

Keeping both files (old and new) will keep support for versions older than Spring Boot 3 and will bring support of newer versions.

Also added documentation on @AfterBotRegistration annotation

### DIFF
--- a/TelegramBots.wiki/FAQ.md
+++ b/TelegramBots.wiki/FAQ.md
@@ -346,3 +346,20 @@ After that your bot will look like:
   }
 ```
 Also you could just implement LongPollingBot or WebHookBot interfaces. All this bots will be registered in context and connected to Telegram api.
+
+In addition, you may use @AfterBotRegistration annotation on one of the methods inside your bot. This method will be called once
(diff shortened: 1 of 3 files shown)


Juratbek:
I am also facing trouble while using this library with Spring Boot 3. Please release a new version for support

equeiroz1:
There is already a PR fixing this: #1196 . I don't know why it's not merged.

@rubenlagus could you please do it ? Thanks in advance.

mikhail-sidnev:
@Juratbek, while auto-config does not work automatically we have a workaround: 
```java
@Import(org.telegram.telegrambots.starter.TelegramBotStarterConfiguration.class)
```

rubenlagus:
This should be fully supported now

PR #1302, opened 2024-01-22
JsonProperty fix for Webhook

Georglider:
Annotated fields with missing JsonProperty in `org/telegram/telegrambots/meta/api/methods` the way it is done in other classes in the same package.
It helps Webhook to return a correctly serialized json result.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,18 +27,18 @@ Just import add the library to your project with one of these options:
     <dependency>
         <groupId>org.telegram</groupId>
         <artifactId>telegrambots</artifactId>
-        <version>6.9.7.0</version>
+        <version>6.9.7.1</version>
     </dependency>
 ```
 
   2. Using Gradle: 
 
 ```gradle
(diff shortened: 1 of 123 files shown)


rubenlagus:
How will you be returning the images/documents in the json?

Georglider:
According to [Telegram API Docs](https://core.telegram.org/bots/api#sending-files), you can send InputFile using its `file_id` or its URL. This works well in a Webhook-designed bot. I encountered this exact issue when was trying to send a photo and MediaGroups this way, so after I made the commit allowing me to return correctly serialized json, my problem was finally fixed and I could finally put a direct link into the InputFile. After that, Telegram successfully responded with a message containing the image or album.

CodeFlusher:
GFH-b60vaPveK-KrFNzFGPHvLEPX13LsA2MtRjEGwqUitz223ffObIc7eu6c-Vtpd5W2HL38WdhMN8xl3d2x2WIH

PR #1319, opened 2024-02-10
Ability to rewrite bots session options on registering bots impl with custom session type

pntrtr:
There is no ability to extend DefaultBotSession class and to set some custom options without access to private options field. For example For example ApiConstants.GETUPDATES_TIMEOUT = 50 is too much for bot.stop() usage and this httpclient timeout can be changed only with protected access for options fiels.

Example usage i want to have:

```
public class TgBotSession extends DefaultBotSession {
    public TgBotSession() {
        this.options = new DefaultBotOptions();
        this.options.setGetUpdatesTimeout(1);
    }

    @Override
    public void setOptions(BotOptions options) {
       // do nothing will ignore options instance from bot impl
    }
}
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+cd telegrambots && mvn clean  install -Dgpg.skip -DskipTests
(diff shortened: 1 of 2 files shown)


rubenlagus:
Bot Options are gone in the new version

PR #1320, opened 2024-02-11
Add official telegram test environment support

mschurov:
Add "/test/" into telegram api url for test environments.

From the docs:

https://core.telegram.org/bots/webapps
****
Using bots in the test environment
To log in to the test environment, use either of the following:

iOS: tap 10 times on the Settings icon > Accounts > Login to another account > Test. 
Telegram Desktop: open ☰ Settings > Shift + Alt + Right click ‘Add Account’ and select ‘Test Server’. 
macOS: click the Settings icon 10 times to open the Debug Menu, ⌘ + click ‘Add Account’ and log in via phone number. 
The test environment is completely separate from the main environment, so you will need to create a new user account and a new bot with @BotFather.

After receiving your bot token, you can send requests to the Bot API in this format:

https://api.telegram.org/bot<token>/**test**/METHOD_NAME
 Note: When working with the test environment, you may use HTTP links without TLS to test your Web App. *****

### DIFF
--- a/telegrambots/src/main/java/org/telegram/telegrambots/bots/DefaultAbsSender.java
+++ b/telegrambots/src/main/java/org/telegram/telegrambots/bots/DefaultAbsSender.java
@@ -25,10 +25,7 @@ import org.telegram.telegrambots.meta.api.methods.send.SendSticker;
 import org.telegram.telegrambots.meta.api.methods.send.SendVideo;
 import org.telegram.telegrambots.meta.api.methods.send.SendVideoNote;
 import org.telegram.telegrambots.meta.api.methods.send.SendVoice;
-import org.telegram.telegrambots.meta.api.methods.stickers.AddStickerToSet;
-import org.telegram.telegrambots.meta.api.methods.stickers.CreateNewStickerSet;
-import org.telegram.telegrambots.meta.api.methods.stickers.SetStickerSetThumb;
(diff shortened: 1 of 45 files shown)


telegrambots/src/main/java/org/telegram/telegrambots/updatesreceivers/DefaultBotSession.java
```diff
@@ -279,6 +279,14 @@ private List<Update> getUpdatesFromServer() throws IOException {
         }
     }
 
+    private String getApiUrl() {
```
4repaXXX, line 282: Why we need this method? It look like duplication of logic with getBaseUrl()

telegrambots/src/main/java/org/telegram/telegrambots/bots/DefaultAbsSender.java
```diff
@@ -123,7 +120,11 @@ public final DefaultBotOptions getOptions() {
     }
 
     public String getBaseUrl() {
-        return options.getBaseUrl() + getBotToken() + "/";
+        String baseUrl = options.getBaseUrl() + getBotToken() + "/";
```
4repaXXX, line 123: Does not it better to allow user set url for http request on start? For example, using standard  setBaseUrl, and just use everywhere method getBaseUrl(). And don't add implicit changes. What if in future telegram decide to change url again.

PR #1326, opened 2024-02-25
Refactored verbose code constructs and QoL improvements

Cespito:
**Summary**
This pull request makes the following improvements:
- Refactor verbose code constructs
- Fix raw use of parameterized class
- Fix redundant type cast
- Add integrety check and small improvements in testing
- Refactor package name
- Delete unnecessary import
- Fix MD list items numbering

**Details**
- Added lambda expressions and utilized library methods to reduce code verbosity
- Such raw use of generic types is valid in Java, but it defeats the purpose of type parameters and may mask bugs
- Removed unnecessary cast expressions
- Added integrety check in SilentSenderTest.java, and replaced in all tests the assertTrue method with check on instanceof with the respective assertInstanceOf library method
- Refactored the org.telegram.telegrambots.test.Fakes package to match the convention. Package names are written in all lower case to avoid conflict with the names of classes or interfaces
- Removed import lombok.extern.jackson.Jacksonized from AnswerInlineQuery.java class as it is not used
- Small fixes on MD files for incorrect numbering in lists

**Benefits**
Improved code organization, maintainability, and readability.
Reduced redundancy and enhanced robustness of the code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,18 +27,18 @@ Just import add the library to your project with one of these options:
     <dependency>
         <groupId>org.telegram</groupId>
         <artifactId>telegrambots</artifactId>
-        <version>6.9.7.0</version>
+        <version>6.9.7.1</version>
     </dependency>
 ```
 
   2. Using Gradle: 
 
 ```gradle
-    implementation 'org.telegram:telegrambots:6.9.7.0'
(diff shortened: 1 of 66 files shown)


rubenlagus:
This is against pre-7 version

PR #1355, opened 2024-05-02
ref(abilities): downgrade log level to debug on `#consume()`

### LINKED ISSUE #1241: abilities: an option to disable logging
**Is your feature request related to a problem? Please describe.**
My app has it's own logging system and handles Telegram updates in it's own way. Currently I'm migrating commands to the abilities system. But the thing is, I don't want it to log any unnecessary stuff.
```java
@Override public void onUpdateReceived(final Update update) {
    super.onUpdateReceived(update); /* call the abilities code */
    /* other logic ... */
}
```

The logging code that I'm reffering to:

https://github.com/rubenlagus/TelegramBots/blob/7029a35cd01dae7f86428ad73e29a4a1d4b725e2/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/bot/BaseAbilityBot.java#L236-L263

**Describe the solution you'd like**
Either one of these would solve the problem; we can just discuss which one suits the library the best and I'll me glad to make a PR implementing the thing.
- extract `Stream.of(...)...` code to a new `protected` `executeAbilities` _(?)_ method;
- elevate access level of other `this::...` `package-protected` methods to `protected`, maybe mark as `final`;
- a toggle on/off for all log calls — it's rather a dirty hack that can be used in place of log filtering.

Disputable:
- don't serialize nulls in `update.toString()` or use Gson for that purpose — that's my main concern, it makes logs completely unreadable.
https://github.com/rubenlagus/TelegramBots/blob/7029a35cd01dae7f86428ad73e29a4a1d4b725e2/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/bot/BaseAbilityBot.java#L238
- log at `trace` level or add a selection for level.

**Describe alternatives you've considered**
Theoretically, in the overriden `#onUpdateReceived()` instead of calling `super.onUpdateReceived(...)` I could just paste the `Stream.of(...` code, but most of the methods are package-protected.
```java
Stream.of(update)
        .filter(this::checkGlobalFlags) /* the only one with protected access level */
        /* ... */
        .forEach(this::postConsumption);
```

A temporary fix is log filtering, but in that case my inner optimization demon will keep me awake at nights telling me that somewhere code's doing unnecessary work :rofl: 
With logback it can be done changing log level for the `...BaseAbilityBot` logger:
```xml
<logger name="org.telegram.abilitybots.api.bot.BaseAbilityBot" level="WARN"/>
```

voidpointer0x00:
fix #1241

### DIFF
--- a/telegrambots-abilities/src/main/java/org/telegram/telegrambots/abilitybots/api/bot/BaseAbilityBot.java
+++ b/telegrambots-abilities/src/main/java/org/telegram/telegrambots/abilitybots/api/bot/BaseAbilityBot.java
@@ -179,8 +179,8 @@ public abstract class BaseAbilityBot implements AbilityExtension, LongPollingSin
      */
     @Override
     public void consume(Update update) {
-        log.info(format("[%s] New update [%s] received at %s", botUsername, update.getUpdateId(), now()));
-        log.info(update.toString());
+        log.debug(format("[%s] New update [%s] received at %s", botUsername, update.getUpdateId(), now()));
+        log.debug(update.toString());
         long millisStarted = System.currentTimeMillis();
 
         Stream.of(update)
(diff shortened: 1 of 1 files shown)


Chase22 (changes requested):
Implementation looks good. Please fix the git foo

voidpointer0x00:
Just rebased on top the upstream, that fixed the conflicts

Chase22 (approved):

PR #1382, opened 2024-06-16
Add common method getType to ReactionType interface

aNNiMON:


Chase22:
Not a big fan of making a generated override for an interface method. The getter is generated in the subtypes using Lombok. I'd argue for explicitly implementing the getter in this case to make the inheritance clearer @rubenlagus your opinion?

aNNiMON:
@Chase22, this change helps to quickly check a type of reaction by using a getter instead of ugly `instanceof`.  
`property = "type"` => `getType()` https://github.com/rubenlagus/TelegramBots/blob/7d502760fb8ad33889ffc3f00c02ac279de4a9c7/telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/objects/reactions/ReactionType.java#L16-L24 

Literally the same was done here: `property = "status"` => `getStatus()` https://github.com/rubenlagus/TelegramBots/blob/7d502760fb8ad33889ffc3f00c02ac279de4a9c7/telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/objects/chatmember/ChatMember.java#L18-L28

And here: https://github.com/rubenlagus/TelegramBots/blob/7d502760fb8ad33889ffc3f00c02ac279de4a9c7/telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/objects/menubutton/MenuButton.java#L34-L50

rubenlagus (approved):

rubenlagus:
Method is already available

PR #1399, opened 2024-07-09
Add optional 'isDebugMode' parameter to AbilityBot implementations

### LINKED ISSUE #886: BaseAbilityBot.runSilently does not provide information on what exactly went wrong
https://github.com/rubenlagus/TelegramBots/blob/4882040822d202e146471d6196e3a40babbdb51a/telegrambots-abilities/src/main/java/org/telegram/abilitybots/api/bot/BaseAbilityBot.java#L664

When handling replies,I often get an error with a message: "_Reply [null] failed to check for conditions. Make sure you're safeguarding against all possible updates._".

This message does not provide any useful information on what happened.
I suggest adding also an error message and the stack to the logged info. So the message could provide more valuable information for solving the issue.

Kr1SEl:
Resolve #886: Add optional 'isDebugMode' parameter to AbilityBot

This change adds a new optional parameter 'isDebugMode' to the AbilityBot class, allowing the method runSilently to enable enhanced logging.


Kr1SEl:
Opened PR to the wrong branch

PR #1400, opened 2024-07-09
Add optional 'isDebugMode' parameter to AbilityBot implementations

Kr1SEl:
Resolve https://github.com/rubenlagus/TelegramBots/issues/886: Add optional 'isDebugMode' parameter to AbilityBot

This change adds a new optional parameter 'isDebugMode' to the AbilityBot class, allowing the method runSilently to enable enhanced logging.


rubenlagus:
Why not just enable debug level logging in the logging framework?

Kr1SEl:
In the problem #886 you said - "That can be added to the bot implementer code directly instead of library enforcing a debug level of logging". I'm not sure if I understand your idea correctly, however I have added this parameter to the bot constructor, so it can be enabled in the implementer code, instead of enforcing library to set a logging level.

Kr1SEl:
I would also like to add the fact that enabling the debug level for the entire log library results in a huge amount of logs being shown in the console, so this parameter will allow the console logs to remain clean.

aNNiMON:
@Kr1SEl you can configure different log levels for each package in almost all loggers. Some references: https://stackoverflow.com/a/47051649/2135221, https://stackoverflow.com/a/47051649/2135221

Kr1SEl:
@aNNiMON Thank you for the reference, it is a good solution, I'm closing a PR

PR #1407, opened 2024-08-04
Dev

rubenlagus:
Dev


telegrambots-client/src/main/java/org/telegram/telegrambots/client/okhttp/OkHttpFutureDownloadCallback.java
```diff
@@ -23,8 +25,10 @@ public void onResponse(@NonNull Call call, @NonNull Response response) {
             if (body == null) {
                 completeExceptionally(new TelegramApiException("Telegram api returned empty response"));
             } else {
-                complete(body.byteStream());
+                complete(new ByteArrayInputStream(IOUtils.toByteArray(body.byteStream())));
```
bluemods, line 28: This would cause OOM for large files, as Telegram allows up to 4GB

rubenlagus: I'm looking to other options in the future but it is required for now to support the method otherwise the stream coming from rest call get closed.

PR #1425, opened 2024-09-07
Jetty HttpClient implementation of TelegramClient

valkuc:
Hi.
In my project I'm using Jetty Server. I came to the conclusion that additional dependencies on OkHttp client and Kotlin - is too much. So I decided to create TelegramClient implementation that uses Jetty HttpClient instead of OkHttp.

To use it, just include 
```
<dependency>
    <groupId>org.telegram</groupId>
    <artifactId>telegrambots-client-jetty-adapter</artifactId>
</dependency>
```
instead of
```
<dependency>
    <groupId>org.telegram</groupId>
    <artifactId>telegrambots-client</artifactId>
</dependency>
```

Some considerations:
1) It would be nice to move OkHttp from telegrambots-client to separate dependency, for example `telegrambots-client-okhttp-adapter`.
2) Move common client parts (like creation of request body) to abstract class.


valkuc:
@rubenlagus Hi. Any troubles with merging?

rubenlagus:
Hi @valkuc , no concerns but can you add one lesson for this client as well here:  https://github.com/rubenlagus/TelegramBotsDocumentation

valkuc:
> Hi @valkuc , no concerns but can you add one lesson for this client as well here: https://github.com/rubenlagus/TelegramBotsDocumentation

Hi. Yep, will do.

Chase22:
> Move common client parts (like creation of request body) to abstract class.

I did actually consider this when making the telegram client, unfortunately there's no unified interface for multipart requests. It's something we'd have to implement ourselves

valkuc:
> > Hi @valkuc , no concerns but can you add one lesson for this client as well here: https://github.com/rubenlagus/TelegramBotsDocumentation
> 
> Hi. Yep, will do.

@rubenlagus maybe I'm blind but I don't see my changes in master. Looks like something goes wrong with merge.

rubenlagus:
> > > Hi @valkuc , no concerns but can you add one lesson for this client as well here: https://github.com/rubenlagus/TelegramBotsDocumentation
> 
> > 
> 
> > Hi. Yep, will do.
> 
> 
> 
> @rubenlagus maybe I'm blind but I don't see my changes in master. Looks like something goes wrong with merge.

I'll merge it when I have a bit of time, there are some extra fields that need to be added to the media classes

PR #1434, opened 2024-09-23
Create <java.version> parameter for pom.xml in all modules

therepanic:
The **<java.version>** parameter for **pom.xml** in all modules has been added. This adds more readability and modifiability in the future. Also, the ability to use this parameter in pom.xml in all modules. Some modules have added this parameter, but some others have not


therepanic:
The changes made in this PR should not affect the build in any way, any idea why Build is breaking?

PR #1446, opened 2024-10-14
Proper handling of errors returned by telegram API

dkBrazz:
Polling Bot 
Errors from telegram were not logged properly - you never know what exacty telegram wants to tell you
All error details are added to the exception message which is always logged with the stacktrace

In Telegram Client HTTP error codes were not hadled at all - `onFailure` is only called on HTTP transport errors (cancellation, a connectivity problem or timeout), HTTP response coded should be handled in `onResponse`
We only knew about TG API errors by seend parsing exceptions, when error structure could not be pased as an expected java class
Now only 2xx responses are parsed normally, for other responses we gather details and throw an exception


Name-not-available:
i haven't check code but seems like a nice addition, is there any way to see what's wrong with building?

telegrambots-client/src/main/java/org/telegram/telegrambots/client/okhttp/AbstractOkHttpFutureCallback.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Override
+    public void onResponse(@NonNull Call call, @NonNull Response response) {
+        if (!response.isSuccessful()) {
+            var messageBuilder = new StringBuilder();
+            messageBuilder.append("code: ").append(response.code());
+            messageBuilder.append(", message: ").append(response.message());
+
+            if(response.body() != null) {
```
Name-not-available, line 30: are you sure you can do it? shouldn't response.body() be called only once? multiple calls will cause an exception

Name-not-available: and all future or past calls too

dkBrazz: It just returns a value of the `body` field
This `if` just a check that body exists and accessing it will not give us an NPE

But the `body` is a `ResponseBody` instance and is consumable and should be consumed once

And here I call `.string()` which consumes body and closes it, so this piece does not require `try-with-resources` notation

dkBrazz:
@rubenlagus any chance to get it merged?

PR #1458, opened 2024-11-19
Api 8.0

rubenlagus:


telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/methods/payments/star/EditUserStarSubscription.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public String getMethod() {
+        return PATH;
+    }
+
+    @Override
+    public void validate() throws TelegramApiValidationException {
+        if (userId == 0) {
+            throw new TelegramApiValidationException("UserId parameters must not be 0", this);
```
aNNiMON: > UserId parameters

`parameter` shouldn't be plural

telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/methods/payments/star/EditUserStarSubscription.java
```diff
@@ -0,0 +1,73 @@
...
+    @NonNull
+    private String telegramPaymentChargeId;
+    /**
+     * Pass True to cancel extension of the user subscription;
+     * the subscription must be active up to the end of the current subscription period.
+     * Pass False to allow the user to re-enable a subscription that was previously canceled by the bot.
+     */
+    @JsonProperty(IS_CANCELLED_FIELD)
+    @NonNull
+    private Boolean isCancelled;
```
aNNiMON: Both `canceled` and `cancelled` are acceptable, but since Bot API [uses single '`l`' character](https://core.telegram.org/bots/api#edituserstarsubscription), isn't is better to stick to official parameter naming?

telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/methods/payments/star/EditUserStarSubscription.java
aNNiMON: There is an inconsistency in the user id validation messages. Probably a constant or even separate validator will be needed in the future to avoid such mistakes. I'll probably provide some PR later.

aNNiMON (approved):

telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/methods/payments/star/EditUserStarSubscription.java
aNNiMON: #1459

telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/methods/payments/star/EditUserStarSubscription.java
rubenlagus: Agree, changed

PR #1459, opened 2024-11-28
Consistency in the common validation messages

aNNiMON:
- fixed inconsistency in the most common validation messages
 - by using `Long` in `requiredUserId` we can detect non-updated `Integer userId` fields at compile time. All of these occurrences replaced with `Long userId`.


aNNiMON:
> Build (TelegramBots) — TeamCity build was canceled by a failing dependency
> Required [Details]`https://localhost:8443/buildConfiguration/TelegramBots_Build/1703`

How am I supposed to know what's causing this problem and do I need to fix something in my PR when CI points to `localhost`?🤯

rubenlagus:
@aNNiMON a build will only happens when I approve it. Until then it will show as failed or pending.

That said, if a build happen is always because the build+test have not been run ahead of raising the MR

telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/methods/SetUserEmojiStatus.java
```diff
@@ -43,7 +44,7 @@ public class SetUserEmojiStatus extends BotApiMethodBoolean {
      */
     @JsonProperty(USER_ID_FIELD)
     @NonNull
-    private Integer userId;
+    private Long userId;
```
rubenlagus, line 47: This is not needed to be a Long, as per docs:

_It is safe to use 32-bit signed integers for storing all Integer fields unless otherwise noted._

telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/methods/gifts/SendGift.java
```diff
@@ -47,7 +48,7 @@ public class SendGift extends BotApiMethodBoolean {
      */
     @JsonProperty(USER_ID_FIELD)
     @NonNull
-    private Integer userId;
+    private Long userId;
```
rubenlagus, line 51: This is not needed to be a Long, as per docs:

_It is safe to use 32-bit signed integers for storing all Integer fields unless otherwise noted._

telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/methods/message/SavePreparedInlineMessage.java
```diff
@@ -45,7 +46,7 @@ public class SavePreparedInlineMessage extends BotApiMethod<PreparedInlineMessag
      */
     @JsonProperty(USER_ID_FIELD)
     @NonNull
-    private Integer userId;
+    private Long userId;
```
rubenlagus, line 49: This is not needed to be a Long, as per docs:

_It is safe to use 32-bit signed integers for storing all Integer fields unless otherwise noted._

telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/methods/payments/star/EditUserStarSubscription.java
```diff
@@ -40,7 +41,7 @@ public class EditUserStarSubscription extends BotApiMethodBoolean {
      */
     @JsonProperty(USER_ID_FIELD)
     @NonNull
-    private Integer userId;
+    private Long userId;
```
rubenlagus, line 44: This is not needed to be a Long, as per docs:

_It is safe to use 32-bit signed integers for storing all Integer fields unless otherwise noted._

telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/methods/SetUserEmojiStatus.java
aNNiMON: @rubenlagus you're absolutely correct. However, `Long` is used everywhere in this library except these 4 new methods

image

I don't see any reason why the same entity `user id` could have different types. For consistency it would be great to use the same type for all user ids. And since `Long` is already used in most places, switching them all back to `Integer` might cause a compatibility problem, but switching `Integer` to `Long` in these 4 new methods won't break anything since the new version hasn't been released yet.

aNNiMON: For user ids it's specified in the docs:

image

rubenlagus (approved):

rubenlagus (approved):

PR #1460, opened 2024-11-30
Bugfix/fix wrong types

khamidjon-khamidov:


telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/methods/message/SavePreparedInlineMessage.java
```diff
@@ -45,7 +45,7 @@ public class SavePreparedInlineMessage extends BotApiMethod<PreparedInlineMessag
      */
     @JsonProperty(USER_ID_FIELD)
     @NonNull
-    private Integer userId;
+    private Long userId;
```
rubenlagus: This is not needed to be a Long, as per docs:

_It is safe to use 32-bit signed integers for storing all Integer fields unless otherwise noted._

khamidjon-khamidov: In my case, it worked after i converted to Long. user id: 7018007084 . I would say it is safe to use Long

rubenlagus (approved):

PR #1461, opened 2024-12-01
bugfix - ignore undefined json keys

khamidjon-khamidov:
I can see you removed  json ignore annotation for some reason. But, Document stopped working for me (not sure about other objects) because Document had some "thumb" key. In general, I suggest leaving the annotation because if new field added at some point to the api, updated objects stops working for released lib versions


rubenlagus:
This is undone in a global way

PR #1466, opened 2024-12-24
BotSession - logs reducing

orthlus:
With long polling and not good network in logs can be a lot of messages with stacktrace. But enough only message, no stacktrace

### DIFF
--- a/telegrambots-longpolling/src/main/java/org/telegram/telegrambots/longpolling/BotSession.java
+++ b/telegrambots-longpolling/src/main/java/org/telegram/telegrambots/longpolling/BotSession.java
@@ -117,7 +117,7 @@ public class BotSession implements AutoCloseable {
                 }
             } catch (TelegramApiErrorResponseException e) {
                 long backOffMillis = backOff.nextBackOffMillis();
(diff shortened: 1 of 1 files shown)


telegrambots-longpolling/src/main/java/org/telegram/telegrambots/longpolling/BotSession.java
```diff
@@ -117,7 +117,7 @@ private ScheduledFuture<?> createPollerTask() throws TelegramApiException {
                 }
             } catch (TelegramApiErrorResponseException e) {
                 long backOffMillis = backOff.nextBackOffMillis();
-                log.error("Error received from Telegram GetUpdates Request, retrying in {} millis...", backOffMillis, e);
+                log.error("Error received from Telegram GetUpdates Request, retrying in {} millis...", backOffMillis);
```
rubenlagus: If not printing the full stacktrace, it can be helpful if we print at least the response from telegram or the actual error why we are retrying

orthlus:
@rubenlagus Hello! I pushed changes in logging exception during longpolling. 
I think it helps with errors like `Caused by: java.net.SocketException: Network unreachable`

orthlus:
@rubenlagus Hello! Sorry for interrupt. Do you need any action from me?

PR #1467, opened 2024-12-26
Fіx possible OOM in downloadFileAsStreamAsync

aNNiMON:
I've been using the `-Xmx80M` JVM option since telegrambots 5.x. Recently I got an OOM error on telegrambots 8.0.0 after trying to download a 17.8M file:

```
Exception in thread "OkHttp Dispatcher" java.lang.OutOfMemoryError: Java heap space
    at org.apache.commons.io.IOUtils.byteArray(IOUtils.java:368)
    at org.apache.commons.io.output.AbstractByteArrayOutputStream.toByteArrayImpl(AbstractByteArrayOutputStream.java:201)                                                             
    at org.apache.commons.io.output.UnsynchronizedByteArrayOutputStream.toByteArray(UnsynchronizedByteArrayOutputStream.java:184)                                                     
    at org.apache.commons.io.IOUtils.toByteArray(IOUtils.java:2703)                      
    at org.telegram.telegrambots.client.okhttp.OkHttpFutureDownloadCallback.onResponse(OkHttpFutureDownloadCallback.java:28)                                                          
```

This PR fixes the issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 
 [![Build Status](https://travis-ci.org/rubenlagus/TelegramBots.svg?branch=master)](https://travis-ci.org/rubenlagus/TelegramBots)
 [![Jitpack](https://jitpack.io/v/rubenlagus/TelegramBots.svg)](https://jitpack.io/#rubenlagus/TelegramBots)
-[![Maven Central](https://maven-badges.herokuapp.com/maven-central/org.telegram/telegrambots-meta/badge.svg)](http://mvnrepository.com/artifact/org.telegram/telegrambots)
+[![Maven Central](https://maven-badges.herokuapp.com/maven-central/org.telegram/telegrambots-meta/badge.svg)](http://mvnrepository.com/artifact/org.telegram/telegrambots-meta)
 [![MIT License](http://img.shields.io/badge/license-MIT-blue.svg?style=flat)](https://github.com/rubenlagus/TelegramBots/blob/master/LICENSE)
 
 A simple to use library to create Telegram Bots in Java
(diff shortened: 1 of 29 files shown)


rubenlagus (approved):

rubenlagus (commented):
Please add a test inside TestTelegramClientIntegration to ensure it works as expected

aNNiMON:
@rubenlagus tests have been added.

Now I see there are two possible usages of `downloadFileAsStreamAsync`:

1. As a callback chain: `client.downloadFileAsStreamAsync(..).thenAccept(is -> /* read file */).then(...)`
2. Get `InputStream` from `completableFuture`: `InputStream is = client.downloadFileAsStreamAsync(...).get()`.

The problem with the second usage is that we cannot get the `InputStream` out of the `completableFuture` without closing the `InputStream`/`responseBody` in the `OkHttpFutureDownloadCallback`.

By making the InputStream buffered, we can safely return a stream from completableFuture:

```java
InputStream is = client.downloadFileAsStreamAsync(...)..thenApply(is -> {
    try {
        return IOUtils.toBufferedInputStream(is);
    } catch (IOException e) {
        throw new RuntimeException(e);
    }
}).get();
```

I've done this for `InputStream downloadFileAsStream(File file)`, but users should do the same if they use `client.downloadFileAsStreamAsync(...).get()`.

----

I also though replacing `complete(body.byteStream());` with `complete(IOUtils.toBufferedInputStream(body.byteStream()));`, but it still consumes additional memory.

Here are my observations for downloading 18_439_895 bytes:

- `complete(new ByteArrayInputStream(IOUtils.toByteArray(body.byteStream())));` -> OOM on  `-Xmx80M`:
- `complete(IOUtils.toBufferedInputStream(body.byteStream()));` -> Works on `-Xmx80M`, OOM on  `-Xmx40M`:
- `complete(body.byteStream());` -> works `-Xmx80M` as well as `-Xmx40M`

PR #1481, opened 2025-01-26
Fix: Remove empty URL validation in SetWebhook to support clearing webhook

wvolfff:
This PR addresses an issue where the SetWebhook class in the library does not allow an empty URL, despite the [Telegram Bot API documentation](https://core.telegram.org/bots/webhooks#:~:text=Both%20parameters%20for%20the%20setWebhook%20method%20are%20classed%20as%20optional.%20Calling%20the%20method%20with%20an%20empty%20URL%20parameter%20can%20be%20used%20to%20clear%20a%20previously%20set%20webhook.) explicitly stating that an empty URL can be used to clear a previously set webhook.

> Both parameters for the setWebhook method are classed as optional. Calling the method with an empty URL parameter can be used to clear a previously set webhook.


rubenlagus:
The documentation states that url is required in [setWebhook](https://core.telegram.org/bots/api#setwebhook), if you need to clear the webhook, you can use [deleteWebhook](https://core.telegram.org/bots/api#deletewebhook) method.

PR #1484, opened 2025-02-05
Adding explicit allowedUpdates

asyncbtd:
1. Created another constructor in the [TelegramBotsLongPollingApplication ](https://github.com/rubenlagus/TelegramBots/blob/master/telegrambots-longpolling/src/main/java/org/telegram/telegrambots/longpolling/TelegramBotsLongPollingApplication.java) class, which additionally accepts a List<String> containing allowedUpdates, which passes allowedUpdates to the [DefaultGetUpdatesGenerator](https://github.com/rubenlagus/TelegramBots/blob/master/telegrambots-longpolling/src/main/java/org/telegram/telegrambots/longpolling/util/DefaultGetUpdatesGenerator.java).
2. In [SpringLongPollingBot ](https://github.com/rubenlagus/TelegramBots/blob/master/telegrambots-springboot-longpolling-starter/src/main/java/org/telegram/telegrambots/longpolling/starter/SpringLongPollingBot.java) added an optional method that sets allowedUpdates in a Spring Boot application. In [TelegramBotInitializer ](https://github.com/rubenlagus/TelegramBots/blob/master/telegrambots-longpolling/src/main/java/org/telegram/telegrambots/longpolling/TelegramBotsLongPollingApplication.java) changed the constructor of the [TelegramBotInitializer ](https://github.com/rubenlagus/TelegramBots/blob/master/telegrambots-springboot-longpolling-starter/src/main/java/org/telegram/telegrambots/longpolling/starter/TelegramBotInitializer.java) class to which the afterPropertiesSet method is accessed.

### DIFF
--- a/telegrambots-longpolling/src/main/java/org/telegram/telegrambots/longpolling/TelegramBotsLongPollingApplication.java
+++ b/telegrambots-longpolling/src/main/java/org/telegram/telegrambots/longpolling/TelegramBotsLongPollingApplication.java
@@ -3,6 +3,7 @@ package org.telegram.telegrambots.longpolling;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import lombok.extern.slf4j.Slf4j;
 import okhttp3.OkHttpClient;
+import org.telegram.telegrambots.longpolling.interfaces.BackOff;
 import org.telegram.telegrambots.longpolling.interfaces.LongPollingUpdateConsumer;
 import org.telegram.telegrambots.longpolling.util.DefaultGetUpdatesGenerator;
 import org.telegram.telegrambots.longpolling.util.ExponentialBackOff;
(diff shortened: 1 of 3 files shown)


telegrambots-longpolling/src/main/java/org/telegram/telegrambots/longpolling/TelegramBotsLongPollingApplication.java
```diff
@@ -63,6 +64,10 @@ public BotSession registerBot(String botToken, LongPollingUpdateConsumer updates
         return registerBot(botToken, () -> TelegramUrl.DEFAULT_URL, new DefaultGetUpdatesGenerator(), updatesConsumer);
     }
 
+    public BotSession registerBot(String botToken, LongPollingUpdateConsumer updatesConsumer, List<String> allowedUpdates) throws TelegramApiException {
```
rubenlagus, line 67: Doesn't it makes more sense to have a constructor like:

`public BotSession registerBot(String botToken, Function<Integer, GetUpdates> getUpdatesGenerator, LongPollingUpdateConsumer updatesConsumer) throws TelegramApiException `

Otherwise the constructor here would grow and grow as more fields of the GetUpdates need to be overriden

asyncbtd: you're right, smarter

asyncbtd (commented):
-

PR #1518, opened 2025-06-20
feat(Message): create `Message#getCommand` method

yvasyliev:
`Message#isCommand` felt lonely, so I added `Message#getCommand` for company


telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/objects/message/Message.java
```diff
@@ -853,15 +865,18 @@ public boolean hasText() {
...
-        }
-        return false;
+        return getCommand() != null;
+    }
+
+    @JsonIgnore
+    public String getCommand() {
+        return Stream.ofNullable(entities)
+                .flatMap(Collection::stream)
+                .filter(entity -> entity != null && NumberUtils.INTEGER_ZERO.equals(entity.getOffset()))
```
rubenlagus: Don't think we need a whole new import to check a number vs 0

yvasyliev: There's nothing wrong with importing constants. It's a bad practice to use magic numbers.

yvasyliev:
Hi @rubenlagus 

1. I replaced the Stream block with plain `for` loop for slightly better performance.
2. `entity.offset` is an `Integer`, meaning just comparing it to `0` is not null-safe. Still `NumberUtils.INTEGER_ZERO` could be handy here (`commons-lang3` is already included into the project). I decided to put `Objects.equals(entity.getOffset(), 0)` instead of `entity.getOffset() != null && entity.getOffset() == 0` if that's fine.

Could you please review the PR?

telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/objects/message/Message.java
```diff
@@ -956,15 +966,26 @@ public boolean hasText() {
...
-            for (MessageEntity entity : entities) {
-                if (entity != null && entity.getOffset() == 0 &&
-                        EntityType.BOTCOMMAND.equals(entity.getType())) {
-                    return true;
+        return getCommand() != null;
+    }
+
+    @JsonIgnore
+    public String getCommand() {
+        if (entities == null) {
```
rubenlagus: We should check for hasText a well

yvasyliev: Added text presence check

yvasyliev:
Hey @rubenlagus could you please review this PR? Thank you!

rubenlagus (approved):

yvasyliev:
The source branch is updated. Kindly requesting you to run the build again

rubenlagus (approved):

PR #1554, opened 2025-12-30
Configurable GetUpdates for SpringLongPollingBot

afds:
Adds Configurable `GetUpdates` for `SpringLongPollingBot`

Otherwise it is impossible to specify what type of allowed updates you want to receive in your `SpringLongPollingBot`.


telegrambots-springboot-longpolling-starter/src/main/java/org/telegram/telegrambots/longpolling/starter/TelegramBotInitializer.java
```diff
@@ -39,7 +38,7 @@ public TelegramBotInitializer(@NonNull TelegramBotsLongPollingApplication telegr
     public void afterPropertiesSet() {
         try {
             for (SpringLongPollingBot longPollingBot : longPollingBots) {
-                BotSession session = telegramBotsApplication.registerBot(longPollingBot.getBotToken(), () -> telegramUrl, new DefaultGetUpdatesGenerator(), longPollingBot.getUpdatesConsumer());
+                BotSession session = telegramBotsApplication.registerBot(longPollingBot.getBotToken(), () -> telegramUrl, longPollingBot::getUpdates, longPollingBot.getUpdatesConsumer());
```
rubenlagus: Any reason why the updates generator is not added as a optional parameter for the TelegramBotInitializer class? Same as the other ones?

PR #1604, opened 2026-08-30
fix: create possibility to attach not uploaded media to rich messages

### LINKED ISSUE #1603: It's not possible to attach not uploaded media when sending rich messages.
**Describe the bug**
It's not possible to attach not uploaded media when sending rich messages. 

[Telegram Bots API documentation](https://core.telegram.org/bots/api#rich-message-formatting-options) says:
When [Markdown-style](https://core.telegram.org/bots/api#rich-markdown-style) or [HTML-style](https://core.telegram.org/bots/api#rich-html-style) formatting is used, you can use links in the form tg://photo?id=..., tg://video?id=..., tg://document?id=..., and tg://audio?id=... instead of an HTTP URL to reuse previously uploaded files or **upload a new file**.

**To Reproduce**
1. Create a InputRichMessage instance.
2. Pass media using InputMediaPhoto, InputMediaAudio etc. and provide unique id of media within of message.
3. Pass html tags with attributes that contains any of these strings:
text:tg://photo?id=
tg://video?id=
 tg://document?id= 
tg://audio?id= 
where is "id" is a unique identifier that were added before.
4. Create a SendRichMessage instance and fill it with other required fields.
5. Execute a request.

**Expected behavior**
Message will sent and media will uploaded.

**Actual behavior**
Telegram responses: [400] Bad Request: can't parse InputRichMessageMedia: media not found

**Additional context**
I looked at the piece of code where the Telegram API is called for this method and saw that there is no check for the presence of new media, so the request is sent in application/json instead of multipart/form-data

minilooth:
Current version creates requests to send rich messages to telegram with application/json content type. This eliminates the possibility to attach not uploaded media to request. I changed behaviour when sending rich messages to use multipart/form-data content type. This PR fixes #1603.

### DIFF
--- a/telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/methods/send/SendRichMessage.java
+++ b/telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/methods/send/SendRichMessage.java
@@ -13,11 +13,13 @@ import lombok.ToString;
 import lombok.experimental.SuperBuilder;
 import lombok.experimental.Tolerate;
 import lombok.extern.jackson.Jacksonized;
-import org.telegram.telegrambots.meta.api.methods.botapimethods.BotApiMethodMessage;
+import org.telegram.telegrambots.meta.api.methods.botapimethods.PartialBotApiMethod;
 import org.telegram.telegrambots.meta.api.objects.ReplyParameters;
+import org.telegram.telegrambots.meta.api.objects.message.Message;
 import org.telegram.telegrambots.meta.api.objects.replykeyboard.ReplyKeyboard;
 import org.telegram.telegrambots.meta.api.objects.richtext.InputRichMessage;
 import org.telegram.telegrambots.meta.api.objects.suggestedpost.SuggestedPostParameters;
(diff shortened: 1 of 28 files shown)


telegrambots-meta/src/main/java/org/telegram/telegrambots/meta/api/methods/send/SendRichMessage.java
```diff
@@ -40,18 +40,18 @@
 public class SendRichMessage extends BotApiMethodMessage {
```
rubenlagus: This should extends from PartialBotApiMethod instead in this case. Like, i.e. SendMediaGroup

minilooth: fixed

rubenlagus:
Change is good to merged once it is covered with UTs

minilooth:
> Change is good to merged once it is covered with UTs

i've added some tests. is it enough ?

minilooth:
@rubenlagus please look

